PR #102, opened 2026-01-08
Change placeholder from 'Host' to 'Username'

Tussky:
Host is ambiguous, you actually mean Username/Email address. I spent about half an hour debugging this, I use my own address I've built and this field only worked for me when I entered in my email address / username. 

Thank you for such a great tool!


andrinoff (approved):
yeah, seems like a good change, thank you for your contribution!

PR #235, opened 2026-03-04
fix: preserve paginated emails during background refresh

### LINKED ISSUE #234: BUG: Paginated emails cannot be opened after background refresh
## Description

Emails loaded via pagination cannot be opened. After scrolling down reasonably far (e.g., past ~50 emails) and loading more via pagination, pressing Enter on one of the newly loaded emails either does nothing or briefly flashes "Fetching email content..." before returning to the inbox.

## Steps to Reproduce

1. Open Matcha with an account that has more than 50 emails
2. Scroll down far enough to trigger pagination (loading emails beyond the initial batch)
3. Try to open one of the newly loaded paginated emails
4. Nothing happens, or a brief loading flash appears

## Expected Behavior

All emails in the list should be openable, regardless of whether they were loaded initially or via pagination.

## Environment

- Matcha version: latest master (commit eda9086)
- Tested with Gmail and Proton Bridge (custom IMAP)
- Reproducible on both providers

Happy to submit a PR for this.

arclayto:
## Summary

- Fix race condition where `EmailsRefreshedMsg` overwrites `m.emailsByAcct` and `m.emails`, discarding emails loaded via pagination

## Root Cause

When the user scrolls down to paginate, `EmailsAppendedMsg` adds new emails to both the inbox's internal list and `m.emails`. However, the background refresh (`EmailsRefreshedMsg`) races with pagination and replaces `m.emailsByAcct` entirely with only the initial batch, wiping paginated emails from `m.emails`. The inbox list still displays them, but pressing Enter fails because the email can no longer be found in `m.emails`.

## Fix

`EmailsRefreshedMsg` now merges refreshed emails with existing paginated emails instead of replacing them. The same fix is applied to the sent mailbox path.

## Test plan

- [x] Scroll past ~50 emails to trigger pagination, open a paginated email — works
- [x] Press `r` to refresh while paginated emails are loaded — paginated emails preserved
- [x] Open emails from initial batch — still works
- [x] Tested on both Gmail and Proton Bridge (custom IMAP)
- [x] `go test ./...` passes

Fixes #234

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/yuin/goldmark v1.7.16
 	github.com/zalando/go-keyring v0.2.6
+	go.mozilla.org/pkcs7 v0.9.0
 	golang.org/x/sys v0.41.0
 	golang.org/x/text v0.34.0
 )
@@ -37,7 +38,6 @@ require (
 	github.com/rivo/uniseg v0.4.7 // indirect
 	github.com/sahilm/fuzzy v0.1.1 // indirect
 	github.com/xo/terminfo v0.0.0-20220910002029-abceb7e1c41e // indirect
-	go.mozilla.org/pkcs7 v0.9.0 // indirect
 	golang.org/x/net v0.47.0 // indirect
 	golang.org/x/sync v0.19.0 // indirect
 )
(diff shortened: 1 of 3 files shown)


andrinoff (changes requested):
There is an issue, that when you open an email that was loaded, it closes after opening

PR #351, opened 2026-03-22
fix: update dialog box wording for draft confirmation

### LINKED ISSUE #345: BUG: Discarding an email does not work
### Describe the bug

When composing an email and then trying to discard it, the email remains in the drafts. This is caused because there's only 2 options when trying to close an email you are composing which are:
- y: This one closes the composing page and saves the email to drafts
- n: This one closes the modal and continues with the email composition

There's nowhere for an option to close the composing page and delete the email.

### To reproduce

1. Start composing an email
2. Discard it
3. Go to drafts and observe the email still being there (note if the drafts were empty before you need to go to some other page and back to the main page as you need to force a re-render for the drafts page to appear)

### Expected behavior

There should be 3 buttons when asking to discard an email:
- y: This should discard the email and delete it entirely and close the compose page
- n: This should save the email in drafts and close the compose page
- c/esc: This should close the modal and return to the email composing(This already exists but there's no visual for it)

### Screenshots

_No response_

### Matcha version

0.27.0

### Go version

_No response_

### OS

Linux mint 22.2

### Terminal

_No response_

### Additional context

_No response_

Marwan051:
## What?

Changed the wording of the dialog box of the leave composer message in composer.go

## Why?

Fixes #345

The old wording was incorrect about discarding the draft. The new wording is clearer about what actually happens to the composed email draft

### DIFF
--- a/tui/composer.go
+++ b/tui/composer.go
@@ -629,7 +629,7 @@ func (m *Composer) View() tea.View {
 	if m.confirmingExit {
 		dialog := DialogBoxStyle.Render(
 			lipgloss.JoinVertical(lipgloss.Center,
-				"Exit email composer? This draft will be saved",
+				"Are you sure you want to exit? This draft will be saved",
 				HelpStyle.Render("\n(y/n)"),
 			),
 		)


tui/composer.go
```diff
@@ -629,7 +629,7 @@ func (m *Composer) View() tea.View {
 	if m.confirmingExit {
 		dialog := DialogBoxStyle.Render(
 			lipgloss.JoinVertical(lipgloss.Center,
-				"Discard draft?",
+				"Leave composer? This draft will be saved",
```
andrinoff: rephrase to "Exit?". keeps it simple

PR #492, opened 2026-04-13
feat: send-as email

mine-13-zoom:
## Summary
- add an account-level `send_as_email` field to keep Gmail OAuth2 login separate from the visible sender address
- use the send-as address for outgoing `From` headers in the composer, SMTP sender, and JMAP sender while leaving auth tied to the login identity
- update docs and add tests covering send-as fallback/display behavior

## Testing
- go test ./...

### DIFF
--- a/backend/jmap/jmap.go
+++ b/backend/jmap/jmap.go
@@ -393,7 +393,7 @@ func (p *Provider) SendEmail(_ context.Context, msg *backend.OutgoingEmail) erro
 
 	// Build raw RFC5322 message and upload as blob
 	var buf bytes.Buffer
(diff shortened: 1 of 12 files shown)


andrinoff:
Hello, @mine-13-zoom ,

Can you, please, describe how it is an issue for you?

Because Gmail's OAuth would be the same as `email` (SMTP login) and `fetch_email` is the one acting as `Send-as`

andrinoff (approved):

mine-13-zoom:
this is incase you have an alias setup for gmail where you send as another email, gmail handles this through that. if you have another domain or a general gmail alias through the same gmail account you can send through this email using this.

PR #525, opened 2026-04-15
fix(imap): close debug log file handle to prevent fd leak

### LINKED ISSUE #504: BUG: Debug IMAP file handle leak
### Describe the bug

In `fetcher/fetcher.go`, when the `DEBUG_IMAP` environment variable is set, a file is opened for debug logging but never closed. Each IMAP connection leaks a file descriptor.

```go
if path := os.Getenv("DEBUG_IMAP"); path != "" {
    if f, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600); err == nil {
        options.DebugWriter = f
    }
}
```

The file `f` is never closed — there is no `defer f.Close()` and no tracking of the handle for later cleanup.

### To reproduce

1. Set `DEBUG_IMAP=/tmp/imap.log`
2. Run matcha and let it make multiple IMAP connections (e.g. fetch emails, IDLE reconnects)
3. Observe file descriptor count growing via `ls /proc/<pid>/fd | wc -l`

### Expected behavior

The debug log file handle should be properly closed when the IMAP connection is closed, or a single shared file handle should be reused across connections.

### OS

All platforms

sueun-dev:
Fixes #504

`connectWithOptions()` was opening a new debug log file handle every time an IMAP connection was created when `DEBUG_IMAP` is set, but never closing it. over time this leaks a file descriptor per connection (fetches, IDLE reconnects, etc).

moved the `os.OpenFile` call into a `sync.Once` so theres one shared handle for the entire process lifetime. the helper `getDebugIMAPWriter()` returns the cached `io.Writer` or nil if the env var isnt set.

pretty straightforward fix, lmk if theres anything I missed

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -16,8 +16,8 @@ import (
 	"net/textproto"
 	"os"
 	"slices"
-	"sync"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/ProtonMail/go-crypto/openpgp"
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -16,6 +16,7 @@ import (
 	"net/textproto"
 	"os"
 	"slices"
+	"sync"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


andrinoff:
@sueun-dev thank you for your contribution!

Please, run `make lint` and commit the changes. Workflow is failing

andrinoff (approved):
lgtm

PR #529, opened 2026-04-16
fix(compose): prioritize reply-to header over from address

### LINKED ISSUE #527: BUG: Mind the Reply-to
### Describe the bug

When replying to an email, the `To` address is set to the `From` address.

### To reproduce

1. Reply to an email, that has an explicit `Reply-to` address

### Expected behavior

The default `To` address should be the `Reply-to` field on the email

### Screenshots

_No response_

### Matcha version

v0.32.0

### Go version

_No response_

### OS

Any

### Terminal

_No response_

### Additional context

_No response_

ronKzl:
## What?

Added `ReplyTo` property tracking across the core data models and protocol parsers.

Added `ReplyTo []string` to `backend.Email` and `fetcher.Email` structs.

Updated `imap`, `jmap`, and `pop3` backend parsers to extract the `Reply-To` header from incoming messages.

Updated `tui.ReplyToEmailMsg` handler in `main.go` to prioritize the `Reply-To` destination (if present) before falling back to the `From` address.

I believe the fix is straightforward but do let me know if there is anything that I missed.

## Why?

Fixes [#527](https://github.com/floatpane/matcha/issues/527)

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -65,6 +65,7 @@ type Email struct {
 	UID         uint32
 	From        string
 	To          []string
+	ReplyTo     []string
 	Subject     string
 	Body        string
 	Date        time.Time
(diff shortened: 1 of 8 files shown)


andrinoff (approved):
lgtm

PR #530, opened 2026-04-16
fix: limit remote image response body to prevent memory exhaustion

### LINKED ISSUE #523: BUG: Remote image fetch has no response body size limit
### Describe the bug

In `view/html.go`, when fetching remote images for HTML email rendering, `io.ReadAll(resp.Body)` reads the entire response body without any size limit. A malicious HTML email could reference a very large image URL (or a URL that streams infinite data slowly), causing excessive memory usage or hanging.

```go
resp, err := client.Get(url)
// ...
data, err := io.ReadAll(resp.Body)
```

While the HTTP client has a 5-second timeout, a slow-drip response within that timeout could still deliver significant data.

### To reproduce

1. Receive an HTML email with an `` tag pointing to a URL that serves a very large file
2. Open the email in matcha
3. Memory usage spikes as the entire response is read into memory

### Expected behavior

Use `io.LimitReader(resp.Body, maxSize)` to cap the response body (e.g. 10MB), similar to how the plugin HTTP API limits responses to 1MB.

### OS

All platforms

Yanhu007:
Fixes #523

## Problem

`io.ReadAll(resp.Body)` reads the entire response without any size limit when fetching remote images for HTML email rendering. A malicious email could reference a URL serving a very large file or slowly streaming infinite data, causing excessive memory usage.

## Fix

Wrap `resp.Body` with `io.LimitReader` capped at 10 MB. This prevents unbounded memory allocation while still allowing reasonable image sizes.

### DIFF
--- a/view/html.go
+++ b/view/html.go
@@ -299,7 +299,10 @@ func fetchRemoteBase64(url string) string {
 		debugImageProtocol("remote fetch non-200 url=%s status=%d", url, resp.StatusCode)
 		return ""
 	}
-	data, err := io.ReadAll(resp.Body)
+	// Limit response body to 10 MB to prevent memory exhaustion from
+	// malicious or very large images.
+	const maxImageSize = 10 << 20 // 10 MB
+	data, err := io.ReadAll(io.LimitReader(resp.Body, maxImageSize))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


andrinoff (approved):
lgtm

Issue #522, opened 2026-04-15
BUG: Desktop notifications fail silently when notify-send/osascript unavailable


andrinoff:
### Describe the bug

In `notify/notify.go`, desktop notifications call `notify-send` (Linux) or `osascript` (macOS) directly. If these tools are not installed (common on minimal Linux installs, Wayland-only setups, or headless systems), the notification silently fails with no feedback to the user.

```go
func Send(title, body string) error {
    switch runtime.GOOS {
    case "darwin":
        script := fmt.Sprintf(\`display notification %q with title %q sound name "default"\`, body, title)
        return exec.Command("osascript", "-e", script).Run()
    case "linux":
        return exec.Command("notify-send", title, body).Run()
    default:
        return nil
    }
}
```

Additionally, on macOS the `fmt.Sprintf` with `%q` should handle escaping, but edge cases with certain Unicode characters could still break the AppleScript.

### To reproduce

1. On a Linux system without `notify-send` installed (or a Wayland session without libnotify)
2. Receive a new email while matcha has notifications enabled
3. No notification appears and no error is shown in the TUI

### Expected behavior

- Check if the notification tool exists before trying to call it
- Show a one-time warning in the TUI if notifications can't be delivered
- Consider supporting D-Bus notifications directly on Linux as an alternative

### OS

Linux (primarily), macOS

floatpanebot:
This has had no activity for 30 days. It will be closed in 14 days unless updated. Comment or remove the `stale` label to keep it open.

floatpanebot:
Closing due to inactivity. Reopen if still relevant.

PR #533, opened 2026-04-16
fix(notify): surface and log errors when notify-send/osascript missing

tmchow:
## Summary

Stop silently losing desktop notifications when `notify-send` / `osascript` aren't installed. `notify.Send` now probes for the helper binary with `exec.LookPath` and returns a descriptive error, and the single production call site (`main.go:630`) wraps the goroutine so the error is logged rather than dropped. Addresses #522.

## Why

`notify.Send` shells out to platform-specific tools to raise desktop notifications. Two failure modes were both invisible to the user:

1. **Binary missing.** Minimal Linux installs, Wayland-only setups, and headless images regularly ship without `libnotify` (`notify-send`), and macOS containers / automated test hosts often omit `osascript`. `exec.Command(...).Run()` just returned that error silently.
2. **Caller throws the error away.** The only production caller, `main.go:630`, used `go notify.Send(...)` which discards the return value entirely. Even without the missing-binary failure, the user had no chance to see anything the Send path reported.

Net effect: on affected setups new-mail notifications never fired and nothing logged -- matches the report in #522.

## Changes

`notify/notify.go`:

- Each OS branch now calls `exec.LookPath(...)` first. If the required tool is not on PATH, return `fmt.Errorf("desktop notifications require <tool>, which was not found on PATH: %w", err)` naming the missing dependency so the caller can log a useful message.
- The happy path is unchanged once the binary exists.

`main.go:630`:

- Wrap the goroutine so `notify.Send`'s error is piped through the existing `log.Printf` path (used elsewhere in the file for backend/oauth failures). Captures folder and account by value so the message is consistent even if `msg`/`acc` mutate between trigger and send.

## Testing

- Added `notify/notify_test.go`:
  - `TestSend_ReturnsErrorWhenBinaryMissing`: replaces `PATH` with an empty directory, runs `Send`, asserts a non-nil error whose message mentions the platform's expected binary (`osascript` on darwin, `notify-send` on linux). Simulates the minimal-install scenario.
  - `TestSend_ReturnsNilOnUnsupportedGOOS`: preserves the original `nil` return on Windows/other platforms.
- `go test ./notify/...` passes.
- Full `go test ./...` shows a pre-existing failure in `view/TestImageProtocolSupported/No_supported_terminals` that is **not** caused by this patch (reproduced on `master` via `git stash`); otherwise clean.
- `go build ./...` passes (CGO libc warnings in `clib/md4c.c` are pre-existing macOS-only `OFF_MAX` macro redefinition warnings, unrelated).

This contribution was developed with AI assistance (Claude Code).

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/main.go
+++ b/main.go
@@ -627,7 +627,11 @@ func (m *mainModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 					accountName = acc.Email
 				}
 			}
-			go notify.Send("Matcha", fmt.Sprintf("New mail in %s (%s)", msg.FolderName, accountName))
+			go func(folderName, accName string) {
+				if err := notify.Send("Matcha", fmt.Sprintf("New mail in %s (%s)", folderName, accName)); err != nil {
+					log.Printf("Desktop notification failed: %v", err)
+				}
+			}(msg.FolderName, accountName)
 		}
 
 		// IDLE detected new mail — refetch the folder if we're viewing it
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -8,12 +8,23 @@ import (
 
 // Send delivers a desktop notification with the given title and body.
 // On macOS it uses osascript; on Linux it uses notify-send.
+//
+// If the platform-specific helper binary is not on PATH (common on minimal
+// Linux installs or Wayland-only setups that do not ship notify-send), Send
+// returns an explicit error instead of silently swallowing the failure so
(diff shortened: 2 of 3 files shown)


andrinoff:
automated agent PR, or written by an AI. 

Closing, against the Contributing Guide

PR #535, opened 2026-04-16
fix(settings): correct crypto config UI/UX

TANUJ-args:
## Summary
- fix crypto configuration key handling so Enter advances through fields and Save/Cancel actions trigger correctly
- fix Space handling so toggle controls respond reliably
- remove unintended horizontal drift in crypto rows by preventing multiline style padding side effects
- keep Save and Cancel on a stable left-aligned row

## Validation
- go test ./...

### DIFF
--- a/tui/settings.go
+++ b/tui/settings.go
@@ -435,6 +435,15 @@ func (m *Settings) updateSMIMEConfig(msg tea.KeyPressMsg) (*Settings, tea.Cmd) {
 			}
 		}
 
+		// Skip Yubikey PIN field when key source is "file"
+		if m.focusIndex == 6 && m.pgpKeySource != "yubikey" {
+			if msg.String() == "shift+tab" || msg.String() == "up" {
+				m.focusIndex = 5
+			} else {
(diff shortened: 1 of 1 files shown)


tui/settings.go
```diff
@@ -408,95 +435,71 @@ func (m *Settings) updateSMIMEConfig(msg tea.KeyPressMsg) (*Settings, tea.Cmd) {
...
+			cmds = append(cmds, setFocus(5))
+			return m, tea.Batch(cmds...)
+		case 5: // PGP key source - enter advances
+			nextFocus := 7
+			if m.pgpKeySource == "yubikey" {
+				nextFocus = 6
 			}
+			cmds = append(cmds, setFocus(nextFocus))
+			return m, tea.Batch(cmds...)
+		case 6: // PGP PIN input - enter advances
```
andrinoff: I know, that this part was faulty before your fix, but while you're at it, you can fix the issue, where it can still focus on the 6th field (Yubikey PIN), even though, the `pgpKeySource` is `"file"`.

PR #606, opened 2026-04-16
fix: email truncation to preserve domain and enforce max width

### LINKED ISSUE #545: BUG: Improve email truncation logic
### Describe the bug
The `truncateEmail` function in `tui/inbox.go` only truncates the local part of the email address if it's longer than 8 characters. This can lead to very long email strings in the "ALL" view account indicators.

### To reproduce
1. Use an account with a long local part or a long domain.
2. View emails in "ALL" view.
3. Observe the account indicator truncation.

### Expected behavior
Better truncation that balances local part and domain, or fits within a fixed width more gracefully.

### Additional context
```go
func truncateEmail(email string) string {
	parts := strings.Split(email, "@")
	if len(parts) >= 1 && len(parts[0]) > 8 {
		return parts[0][:8] + "..."
	}
	if len(parts) >= 1 {
		return parts[0]
	}
	return email
}
```

ritikaa1802:
The current truncateEmail function only truncates the local part and removes the domain.

This fix:
- Preserves both local and domain
- Improves readability

Example:
verylongemail@gmail.com → verylo...@gmail.com

Fixes #545

### DIFF
--- a/tui/inbox.go
+++ b/tui/inbox.go
@@ -211,14 +211,29 @@ func parseSenderName(from string) string {
 
 // truncateEmail shortens an email for display
 func truncateEmail(email string) string {
-	parts := strings.Split(email, "@")
-	if len(parts) >= 1 && len(parts[0]) > 8 {
-		return parts[0][:8] + "..."
+	maxLength := 18
+
+	if len(email) <= maxLength {
+		return email
 	}
-	if len(parts) >= 1 {
-		return parts[0]
+
+	parts := strings.SplitN(email, "@", 2)
+	if len(parts) != 2 {
+		if len(email) > maxLength {
+			return email[:maxLength-3] + "..."
+		}
+		return email
 	}
-	return email
+
+	local := parts[0]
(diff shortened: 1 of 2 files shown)


ritikaa1802:
Hi, I’ve implemented the fix based on the discussed approach.

Would love your feedback.

tui/inbox.go
```diff
@@ -211,14 +211,29 @@ func parseSenderName(from string) string {
...
+		}
+		return email
 	}
-	return email
+
+	local := parts[0]
+	domain := parts[1]
+
+	// Keep full domain visible (e.g. ...@gmail.com) and truncate local part first.
+	if len(local) > 6 {
```
andrinoff: the issue we were talking about, and your comment on it mentions 8 characters before truncation.

please, change it to 8, because with 6 characters kept, 1/3 of the total local part is `...`

andrinoff (approved):
lgtm

ritikaa1802:
Glad! it helped.

PR #666, opened 2026-04-17
fix: yubikey bounds-check

### LINKED ISSUE #613: BUG: Missing bounds validation in ASN.1 signature parsing
### Describe the bug

In `pgp/yubikey.go` lines 428-440, `parseASN1Signature()` doesn't validate slice bounds before accessing:

```go
rLen := int(der[pos])
pos++
rVal := new(big.Int).SetBytes(der[pos : pos+rLen])  // No bounds check
pos += rLen
```

If `pos + rLen` exceeds `len(der)`, panic occurs (index out of range). Initial check only validates `len(der) < 6`, but not actual slice access.

### To reproduce

1. Send malformed ECDSA signature with large `rLen` value
2. Parser attempts slice access beyond array bounds
3. Application panics with index out of range

### Expected behavior

Add bounds validation before slice access:

```go
rLen := int(der[pos])
pos++
if pos+rLen > len(der) {
    return nil, nil, fmt.Errorf("ASN.1 signature truncated: r length overflow")
}
rVal := new(big.Int).SetBytes(der[pos : pos+rLen])
pos += rLen
```

### Screenshots

N/A

### Additional context

- **File:** `pgp/yubikey.go`
- **Lines:** 428-440
- **Severity:** High - crash on malformed signatures, potential DoS
- **Fix complexity:** Easy - add bounds check before slice operations

SAY-5:
### Problem

`parseASN1Signature` in `pgp/yubikey.go` declares a 6-byte minimum length
up front and then walks `der` using the declared INTEGER length bytes with
no further bounds check:

```go
rLen := int(der[pos])
pos++
rVal := new(big.Int).SetBytes(der[pos : pos+rLen])  // no bounds check
pos += rLen
```

If a signature's declared `rLen` (or `sLen`) exceeds the bytes that
actually follow, the slice expression panics with
`runtime error: index out of range`. For a routine reachable from
attacker-influenced inputs this is a trivial DoS (#613).

### Fix

Bounds-check every intermediate `der[pos]` read and every
`der[pos:pos+N]` slice. A truncated or malformed signature now returns
a typed error rather than panicking.

```diff
-    if der[pos] != 0x02 {
+    if pos >= len(der) || der[pos] != 0x02 {
         return nil, nil, fmt.Errorf("expected INTEGER tag for R")
     }
     pos++
+    if pos >= len(der) {
+        return nil, nil, fmt.Errorf("ASN.1 signature truncated before R length")
+    }
     rLen := int(der[pos])
     pos++
+    if rLen < 0 || pos+rLen > len(der) {
+        return nil, nil, fmt.Errorf("ASN.1 signature truncated: R length overflow")
+    }
     rVal := new(big.Int).SetBytes(der[pos : pos+rLen])
```

Same guards for S. Happy path bytes are unchanged.

Added `pgp/yubikey_test.go` with:

- `R length overruns buffer` — payload that previously panicked
- `S length overruns buffer` — payload that previously panicked
- `missing S after R` — catches the `pos >= len(der)` branch
- `WellFormed` — regression test for the normal path

Fixes #613

### DIFF
--- a/pgp/yubikey.go
+++ b/pgp/yubikey.go
@@ -414,6 +414,10 @@ func writeNewFormatLength(w *bytes.Buffer, length int) {
 }
 
 // parseASN1Signature extracts r and s from an ASN.1 DER encoded ECDSA signature.
+//
+// Each intermediate slice access is bounds-checked against len(der). A truncated
+// or malformed signature produces a typed error rather than an index-out-of-range
+// panic; the minimum-length check up front only rules out obvious runts (#613).
 func parseASN1Signature(der []byte) (r, s []byte, err error) {
 	// ASN.1 SEQUENCE { INTEGER r, INTEGER s }
 	if len(der) < 6 || der[0] != 0x30 {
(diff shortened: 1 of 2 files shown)


pgp/yubikey.go
```diff
@@ -423,22 +427,34 @@ func parseASN1Signature(der []byte) (r, s []byte, err error) {
...
+	if pos >= len(der) || der[pos] != 0x02 {
 		return nil, nil, fmt.Errorf("expected INTEGER tag for R")
 	}
 	pos++
+	if pos >= len(der) {
+		return nil, nil, fmt.Errorf("ASN.1 signature truncated before R length")
+	}
 	rLen := int(der[pos])
 	pos++
+	if rLen < 0 || pos+rLen > len(der) {
```
andrinoff: `rLen` is always >= 0, because `der[pos]` is a `uint8` ([0;255])

andrinoff (changes requested):
Please, fix the issue in the comment i left. @SAY-5

andrinoff (approved):
lgtm

PR #667, opened 2026-04-17
sender: return errors from quoted-printable encoding in SendEmail

### LINKED ISSUE #614: BUG: Unchecked error on quoted-printable writer Close()
### Describe the bug

In `sender/sender.go` line 220, `quotedprintable.NewWriter().Close()` error is ignored:

```go
var encBody bytes.Buffer
qp := quotedprintable.NewWriter(&encBody)
fmt.Fprint(qp, plainBody)
qp.Close()  // Error ignored
encodedBody := encBody.Bytes()
```

If `Close()` fails (write errors, flush failures), encoded body may be incomplete/corrupted but code proceeds silently.

### To reproduce

1. Compose email with plain text body
2. Simulate write error during quoted-printable encoding
3. `Close()` fails but error not caught
4. Corrupted email sent

### Expected behavior

Check `Close()` error:

```go
var encBody bytes.Buffer
qp := quotedprintable.NewWriter(&encBody)
if _, err := fmt.Fprint(qp, plainBody); err != nil {
    return err
}
if err := qp.Close(); err != nil {
    return fmt.Errorf("quoted-printable encoding failed: %w", err)
}
encodedBody := encBody.Bytes()
```

### Screenshots

N/A

### Additional context

- **File:** `sender/sender.go`
- **Lines:** 217-221
- **Severity:** Medium - email body corruption
- **Fix complexity:** Easy - add error check

SAY-5:
### Problem

`sender/sender.go` builds quoted-printable bodies in three places and
drops the `Close` error on the floor every time:

```go
qp := quotedprintable.NewWriter(&encBody)
fmt.Fprint(qp, plainBody)
qp.Close()  // error ignored
```

`quotedprintable.Writer.Close` flushes the final soft-line-break state
and any buffered partial output, so a silent failure there hands the
rest of the MIME assembly a truncated or ill-formed body, and the
corrupted message is sent. #614 flagged the plaintext-only path at
line 220; the same pattern exists twice more in the multipart path:

- line 220: plaintext-only body
- line 360: `text/plain` part inside `multipart/alternative`
- line 373: `text/html` part inside `multipart/alternative`

### Fix

`SendEmail` already returns `([]byte, error)`. Propagate the error from
`Fprint` and `Close` at all three sites with a wrapped message so the
caller can tell which part failed. No behaviour change on the happy path.

```diff
- fmt.Fprint(qp, plainBody)
- qp.Close()
+ if _, err := fmt.Fprint(qp, plainBody); err != nil {
+     return nil, fmt.Errorf("encoding plain body: %w", err)
+ }
+ if err := qp.Close(); err != nil {
+     return nil, fmt.Errorf("encoding plain body: %w", err)
+ }
```

Fixes #614

### DIFF
--- a/sender/sender.go
+++ b/sender/sender.go
@@ -213,11 +213,17 @@ func SendEmail(account *config.Account, to, cc, bcc []string, subject, plainBody
 			return nil, errors.New("plaintext-only messages cannot contain attachments or inline images")
 		}
 
-		// Build quoted-printable encoded body
+		// Build quoted-printable encoded body. Close() flushes the final
+		// soft-line-break state, so an ignored error here would silently ship
+		// a truncated body. See #614.
 		var encBody bytes.Buffer
 		qp := quotedprintable.NewWriter(&encBody)
-		fmt.Fprint(qp, plainBody)
-		qp.Close()
+		if _, err := fmt.Fprint(qp, plainBody); err != nil {
+			return nil, fmt.Errorf("encoding plain body: %w", err)
+		}
+		if err := qp.Close(); err != nil {
+			return nil, fmt.Errorf("encoding plain body: %w", err)
+		}
 		encodedBody := encBody.Bytes()
 
 		// Build the canonical MIME part (headers + body) used for signing/encryption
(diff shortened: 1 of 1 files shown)


andrinoff (changes requested):

PR #668, opened 2026-04-17
fix: log keyring failures

### LINKED ISSUE #616: BUG: Keyring operations silently ignore errors
### Describe the bug

In `config/config.go` lines 357-360 and 365-366, keyring operations suppress errors with `_`:

```go
_ = keyring.Set(keyringServiceName, acc.Email, acc.Password)
_ = keyring.Delete(keyringServiceName, acc.Email)
```

If keyring unavailable or fails, passwords silently fail to persist with no warning/fallback.

### To reproduce

1. Run matcha on system without keyring support
2. Save password - keyring.Set fails silently
3. Restart matcha - password lost, no indication why
4. User confused about missing credentials

### Expected behavior

Log keyring errors as warnings:

```go
if err := keyring.Set(keyringServiceName, acc.Email, acc.Password); err != nil {
    log.Printf("Warning: failed to store password in keyring: %v", err)
}
```

Or provide fallback mechanism for systems without keyring.

### Screenshots

N/A

### Additional context

- **File:** `config/config.go`
- **Lines:** 357-360, 365-366
- **Severity:** Medium - silent credential loss
- **Fix complexity:** Easy - add logging/warning

SAY-5:
### Problem

`config/config.go` has five keyring calls whose errors are swallowed with
`_ =`:

- `SaveConfig`: `keyring.Set` for `acc.Password` and `acc.Password + ":pgp-pin"`
- Legacy password migration: `keyring.Set(rawAcc.Email, rawAcc.Password)`
- `RemoveAccount`: `keyring.Delete` for `acc.Email` and `:pgp-pin`

If the OS keyring is unavailable, locked, or otherwise unhappy, the
password and PGP PIN silently fail to persist. The user is then asked to
reauthenticate on every restart with no indication that keyring storage
ever failed. Same pattern reported in #616.

### Fix

Log a warning on each failure:

```go
if err := keyring.Set(keyringServiceName, acc.Email, acc.Password); err != nil {
    log.Printf("matcha: failed to store password for %s in keyring: %v", acc.Email, err)
}
```

For the two `Delete` paths, swallow `keyring.ErrNotFound` (expected when
the entry never existed or was already cleaned up) and log anything else.

No behaviour change on keyring success. Adds `log` to the import list.

Fixes #616

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"encoding/json"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


andrinoff (approved):
lgtm

PR #670, opened 2026-04-17
pgp/yubikey: propagate write errors from writeMPI

### LINKED ISSUE #612: BUG: Ignored write errors in PGP signature generation (writeMPI)
### Describe the bug

In `pgp/yubikey.go` lines 386-387, the `writeMPI()` function ignores write errors using `//nolint:errcheck`:

```go
w.Write(buf)  //nolint:errcheck
w.Write(data) //nolint:errcheck
```

These write operations could fail silently if buffer capacity is reached or I/O errors occur, producing invalid OpenPGP signatures without error indication.

### To reproduce

1. Trigger PGP signature generation with Yubikey
2. If buffer writes fail (simulated via full buffer), signatures corrupt silently
3. No error returned to caller

### Expected behavior

Write errors should be checked and propagated:

```go
if _, err := w.Write(buf); err != nil {
    return err
}
if _, err := w.Write(data); err != nil {
    return err
}
```

### Screenshots

N/A

### Additional context

- **File:** `pgp/yubikey.go`
- **Lines:** 386-387
- **Severity:** High - silent signature corruption possible
- **Fix complexity:** Easy - add error checks and propagate

SAY-5:
### Problem

`writeMPI` in `pgp/yubikey.go` takes an `io.Writer` and silently drops
both of its `Write` calls with `//nolint:errcheck`:

```go
func writeMPI(w io.Writer, data []byte) {
    ...
    w.Write(buf)  //nolint:errcheck
    w.Write(data) //nolint:errcheck
}
```

Every in-tree caller today passes `&body` (a `*bytes.Buffer`), where
`Write` never fails in practice, so the linter suppression has been
"fine". But the parameter is declared as `io.Writer`, which advertises
support for any writer: an armor writer capped by the ASCII-armor
layer, a network writer that disconnects mid-signature, etc. If any
of those Writers return an error, we currently construct and return a
truncated OpenPGP signature packet with no indication of failure (#612).

### Fix

Change `writeMPI` to return `error` and wrap each `Write` in
`fmt.Errorf` so the failure surfaces as a proper signing error.
Update the three call sites in the YubiKey signing switch to check
and propagate:

```diff
- writeMPI(&body, rawSig[:32]) // r
- writeMPI(&body, rawSig[32:]) // s
+ if err := writeMPI(&body, rawSig[:32]); err != nil {
+     return nil, err
+ }
+ if err := writeMPI(&body, rawSig[32:]); err != nil {
+     return nil, err
+ }
```

Same treatment for the RSA and ECDSA arms. This matches the existing
pattern in the same file for armor-writer writes in `armorSignature`
(line 265).

Fixes #612

### DIFF
--- a/pgp/yubikey.go
+++ b/pgp/yubikey.go
@@ -225,12 +225,18 @@ func buildSignaturePacket(signedContent []byte, signer crypto.Signer, pubKey *pa
 		if len(rawSig) != 64 {
 			return nil, fmt.Errorf("unexpected EdDSA signature length: %d", len(rawSig))
 		}
-		writeMPI(&body, rawSig[:32]) // r
-		writeMPI(&body, rawSig[32:]) // s
+		if err := writeMPI(&body, rawSig[:32]); err != nil {
+			return nil, err
+		}
+		if err := writeMPI(&body, rawSig[32:]); err != nil {
+			return nil, err
+		}
 
 	case packet.PubKeyAlgoRSA, packet.PubKeyAlgoRSASignOnly:
 		// RSA: single MPI
-		writeMPI(&body, rawSig)
+		if err := writeMPI(&body, rawSig); err != nil {
+			return nil, err
+		}
 
 	case packet.PubKeyAlgoECDSA:
 		// ECDSA: card returns ASN.1 DER encoded (R, S)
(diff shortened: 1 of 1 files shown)


andrinoff (changes requested):

PR #671, opened 2026-04-17
main: sanitize attachment filenames before writing to disk

### LINKED ISSUE #642: BUG: Attachment filename not sanitized for filesystem
Describe the bug: Attachment filenames used directly when saving to disk. Could contain path traversal characters like "../" or invalid filesystem chars.

To reproduce:
1. Receive email with attachment named "../../evil.txt"
2. Save attachment
3. May write outside intended directory

Expected behavior: Sanitize filename:
- Strip path separators (/ \)
- Remove invalid chars for current OS
- Limit length to filesystem max
- Handle collisions with (1), (2) suffix

Additional context:
- File: view/html.go or attachment handling code
- Use filepath.Base() and validate chars
- Severity: Medium - path traversal risk
- Complexity: Easy - filename sanitization

SAY-5:
### Problem

`downloadAttachmentCmd` takes `msg.Filename` (the MIME
Content-Disposition filename set by the sender) and passes it directly
into `filepath.Join(downloadsPath, candidate)` inside an
`O_CREATE|O_EXCL` open:

```go
origName := msg.Filename
ext := filepath.Ext(origName)
base := strings.TrimSuffix(origName, ext)
candidate := origName
...
filePath = filepath.Join(downloadsPath, candidate)
f, err := os.OpenFile(filePath, os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0644)
```

The sender controls that header, so a hostile email can send

```
Content-Disposition: attachment; filename="../../etc/passwd"
Content-Disposition: attachment; filename="..\\evil.txt"
```

and the exclusive-create happens outside `~/Downloads`. `O_EXCL` means
the attack needs a path that does not already exist, but placing a file
anywhere the user can write (e.g. `~/.ssh/authorized_keys2`,
`~/.config/<something>`, a shell init dotfile) is still reachable (#642).

### Fix

Add `sanitizeAttachmentFilename` and wire it into `origName`:

- `filepath.Base` to drop any directory prefix (`../../etc/passwd` → `passwd`)
- replace both `/` and `\` so Windows-authored headers can't sneak a
  segment through on unix, and vice versa
- strip NUL bytes
- trim leading `.` / space so `.bashrc` or `...` doesn't silently
  create a hidden file
- fall back to `"attachment"` when the cleaned name is empty

Happy-path filenames (e.g. `report.pdf`) round-trip unchanged. The
`(n)` suffix logic and `downloadsPath` creation are untouched.

Fixes #642

### DIFF
--- a/main.go
+++ b/main.go
@@ -2390,6 +2390,36 @@ func moveEmailToFolderCmd(account *config.Account, uid uint32, accountID string,
 	}
 }
 
+// sanitizeAttachmentFilename returns a filename safe to join with a trusted
+// directory. It strips any directory component the sender put in the name
+// (so "../../etc/passwd" collapses to "passwd"), replaces both path separators
+// even on the wrong platform (Windows clients can see "/", unix clients "\"),
+// trims leading dots that would produce hidden files, and falls back to a
+// generic name when the sender supplies an empty or nothing-but-separators
+// string. Used by the attachment download path; see #642.
+func sanitizeAttachmentFilename(name string) string {
+	// Drop any directory component (e.g. "foo/bar.txt" -> "bar.txt",
+	// "../../etc/passwd" -> "passwd"). filepath.Base handles the active
+	// OS separator; we also strip the alternate separator so a Windows
+	// name on a unix host (or vice versa) cannot sneak a segment through.
+	name = filepath.Base(name)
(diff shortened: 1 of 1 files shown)


andrinoff (changes requested):

Issue #519, opened 2026-04-15
BUG: Account deletion leaves orphaned cache files


andrinoff:
### Describe the bug

When an account is deleted via settings, `config.RemoveAccount()` only removes the account from the config array and deletes the password from the OS keyring. It does not clean up:

- Cached emails in `~/.cache/matcha/folder_emails/`
- Cached email bodies in `~/.cache/matcha/email_bodies/`
- Contacts associated with that account
- Any draft data for that account

Over time, this accumulates stale data on disk.

### To reproduce

1. Add an account and fetch emails
2. Delete the account from settings
3. Check `~/.cache/matcha/` — cached data for the deleted account remains

### Expected behavior

Deleting an account should also clean up all associated cache files, or at minimum prompt the user to optionally delete cached data.

### OS

All platforms

PR #672, opened 2026-04-18
config: drop an account's drafts when it is removed

### LINKED ISSUE #566: BUG: Orphaned drafts after account removal
### Describe the bug
When an account is removed via `config.RemoveAccount`, its associated drafts are still kept in the `drafts.json` cache.

### To reproduce
1. Create a draft for Account A.
2. Remove Account A.
3. Observe that the draft for Account A still exists in the drafts list (though it might fail to load properly).

### Expected behavior
Drafts associated with a removed account should be cleaned up or handled gracefully.

### Additional context
Similar to #519.

SAY-5:
### Problem

`Config.RemoveAccount` drops the account from `c.Accounts` and clears
its keyring entries, but leaves `drafts.json` untouched. Every draft
the user wrote under that account stays in the cache keyed by an
`AccountID` that now points at nothing. The drafts pane shows drafts
that cannot be loaded or sent, and if the user later adds a new
account, stale rows with matching IDs can collide. #566.

### Fix

- Add `DeleteDraftsForAccount(accountID string) error` next to the
  other draft helpers in `config/cache.go`. It loads the drafts cache,
  filters out drafts whose `AccountID` matches, and writes the cache
  back. No-ops when the cache is absent or when nothing changes (so we
  don't churn `drafts.json` for accounts that never had a draft).
- Call it from `RemoveAccount` right after the keyring cleanup and
  before the `append` that removes the account from the slice.

```diff
  _ = keyring.Delete(keyringServiceName, acc.Email)
  _ = keyring.Delete(keyringServiceName, acc.Email+":pgp-pin")

+ // Drop drafts tied to the removed account so they do not
+ // linger in drafts.json referencing a now-unknown AccountID
+ // (#566).
+ _ = DeleteDraftsForAccount(id)

  c.Accounts = append(c.Accounts[:i], c.Accounts[i+1:]...)
```

Draft deletion errors are swallowed the same way the keyring cleanup
errors are: removing the account succeeds even if cache cleanup fails.
(#519 tracks the broader cache-orphan story; this PR addresses drafts
specifically.)

Fixes #566

### DIFF
--- a/config/cache.go
+++ b/config/cache.go
@@ -363,6 +363,32 @@ func DeleteDraft(id string) error {
 	return SaveDraftsCache(cache)
 }
 
+// DeleteDraftsForAccount removes every draft belonging to the given account ID
+// from the drafts cache. Used when an account is removed via RemoveAccount so
+// drafts for the removed account do not linger in the cache as orphans (#566).
+//
+// No-op if the cache does not exist or contains no drafts for the account.
+func DeleteDraftsForAccount(accountID string) error {
+	cache, err := LoadDraftsCache()
+	if err != nil {
+		return nil // No cache, nothing to delete
+	}
+
+	var filtered []Draft
+	for _, d := range cache.Drafts {
+		if d.AccountID != accountID {
+			filtered = append(filtered, d)
+		}
+	}
+	if len(filtered) == len(cache.Drafts) {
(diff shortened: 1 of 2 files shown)


andrinoff (changes requested):

PR #674, opened 2026-04-18
config/encryption: use crypto/subtle for sentinel comparison

### LINKED ISSUE #659: BUG: Encryption key derivation vulnerable to timing attacks
Describe the bug: Password comparison in encryption code may leak timing information. Allows offline timing attack on encrypted data to guess password.

To reproduce:
1. Capture encrypted config file
2. Script password guessing with timing measurement
3. Timing variations reveal password characters
4. Brute force becomes more efficient

Expected behavior:
- Use constant-time comparison (subtle.ConstantTimeCompare)
- Ensure Argon2 parameters prevent timing leaks
- Add rate limiting to password attempts
- Consider authentication tag before decryption

Additional context:
- File: config/encryption.go
- Use crypto/subtle for comparisons
- Review entire auth flow for timing leaks
- Severity: Medium - offline attack vector
- Complexity: Medium - cryptography expertise needed
- Related: CWE-208 timing side channel

SAY-5:
### Problem

`VerifyPassword` in `config/encryption.go` decrypts a fixed sentinel
string with the key derived from the user's password and then compares
the plaintext against the constant `sentinelPlaintext` with a plain
`!=`:

```go
if string(plaintext) != sentinelPlaintext {
    return nil, errors.New("incorrect password")
}
```

Go's `string !=` short-circuits on the first differing byte, and also
on any length mismatch before it starts comparing. The amount of time
`VerifyPassword` takes therefore varies with whatever plaintext the
attacker managed to produce — a classic CWE-208 timing side-channel
against an offline attacker who already has a copy of `secure.meta`.
#659.

### Fix

Swap in `subtle.ConstantTimeCompare` against a `[]byte`:

```diff
- if string(plaintext) != sentinelPlaintext {
+ if subtle.ConstantTimeCompare(plaintext, []byte(sentinelPlaintext)) != 1 {
      return nil, errors.New("incorrect password")
  }
```

Adds `crypto/subtle` to the imports. The common wrong-key case is
already caught by the `Decrypt` error branch above this block and
doesn't touch the compare, so the happy path and the common failure
path are both unaffected.

Fixes #659

### DIFF
--- a/config/encryption.go
+++ b/config/encryption.go
@@ -4,6 +4,7 @@ import (
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/rand"
+	"crypto/subtle"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


andrinoff (changes requested):

PR #676, opened 2026-04-18
fix: validate PGPKeySource field on config load

### LINKED ISSUE #620: BUG: Missing PGPKeySource validation in config
Describe the bug: PGPKeySource field accepts any string but no validation when loading config.

To reproduce:
1. Edit config.json, set pgp_key_source: "invalid"
2. Load config - no error
3. Later code assumes valid value

Expected behavior: Validate on config load to ensure value is "file" or "yubikey" only.

Additional context:
- File: config/config.go
- Severity: Low
- Fix: Add validation function

mvanhorn:
PGPKeySource accepted any string without validation. An invalid value like `pgp_key_source: "invalid"` would load successfully but cause silent failures when the code later assumes the value is `"file"` or `"yubikey"`.

Added validation in `LoadConfig()` immediately after the account struct is constructed. Now rejects invalid values with a clear error message at startup rather than failing silently at runtime.

Closes #620

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -565,6 +565,11 @@ func LoadConfig() (*Config, error) {
 			POP3Port:           rawAcc.POP3Port,
 		}
 
+		// Validate PGPKeySource
+		if acc.PGPKeySource != "" && acc.PGPKeySource != "file" && acc.PGPKeySource != "yubikey" {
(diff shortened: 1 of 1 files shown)


andrinoff (approved):
lgtm

PR #712, opened 2026-04-20
fix(fetcher): replace fmt.Print with standard logger

### LINKED ISSUE #702: BUG: Hardcoded fmt.Print in fetcher
### Describe the bug
In `fetcher/fetcher.go:1025`, `fmt.Print(msg)` is used instead of a standard logger.

### To reproduce
Trigger the specific fetcher code path.

### Expected behavior
Use the application's standard logging facility.

Haroka-74:
## Description

Replaced a raw `fmt.Print` call in the fetcher with the standard logging.

Closes #702

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"mime"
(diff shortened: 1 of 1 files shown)


andrinoff (approved):
lgtm

PR #748, opened 2026-04-21
fix: use atomic increment for image ID allocation to prevent race condition

### LINKED ISSUE #730: BUG: Race condition in image ID allocation (view/html.go)
**Describe the bug**
In view/html.go, nextImageID is incremented without mutex protection, causing potential race conditions when multiple goroutines process images concurrently.

**To reproduce**
Search view/html.go for nextImageID usage - if it's a package-level variable incremented without sync protection, concurrent access causes races.

**Expected behavior**
Use atomic.AddUint32() or protect with mutex.

**Additional context**
- Good first issue if confirmed
- Race detector would catch this: `go test -race`
- Verify with: `grep -n "nextImageID" view/html.go`

**Suggested fix:**
```go
import "sync/atomic"

var nextImageID uint32

// In code:
id := atomic.AddUint32(&nextImageID, 1)
```

**OS**
All platforms

**Note:** Needs verification - issue based on common pattern analysis.

kyooosukedn:
## What?

Replaced the non-atomic `nextImageID++` increment in `allocImageID()` with `atomic.AddUint32(&nextImageID, 1)`.

## Why?

Fixes #730

`allocImageID()` in `view/html.go` used a plain read-then-increment (`id := nextImageID; nextImageID++`) with no synchronization. When multiple goroutines process images concurrently (e.g., rendering an email with multiple inline images), they can:

1. Read the same value of `nextImageID`, allocating duplicate IDs
2. Lose updates entirely due to the non-atomic write

This would cause Kitty image protocol placements to conflict, resulting in broken image rendering. The race detector (`go test -race`) would flag this immediately.

The fix uses `sync/atomic.AddUint32` which is a single CPU instruction on most architectures — zero overhead.

## Testing

Added `TestAllocImageID_NoRace` which:
- Spawns 100 goroutines, each allocating 100 IDs concurrently
- Verifies all 10,000 IDs are unique (no duplicates)
- Would deterministically fail under the old code with `go test -race`

### DIFF
--- a/view/html.go
+++ b/view/html.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"charm.land/lipgloss/v2"
@@ -282,9 +283,7 @@ var nextImageID uint32 = 1000
 
 // allocImageID returns a unique Kitty image ID.
 func allocImageID() uint32 {
-	id := nextImageID
-	nextImageID++
-	return id
+	return atomic.AddUint32(&nextImageID, 1)
 }
 
 func fetchRemoteBase64(url string) string {
--- a/view/html_test.go
+++ b/view/html_test.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"regexp"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"testing"
 
 	"charm.land/lipgloss/v2"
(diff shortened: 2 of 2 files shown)


andrinoff (approved):

andrinoff:
lgtm

floatpanebot (approved):

PR #749, opened 2026-04-21
fix: use local time, and highlight dates

### LINKED ISSUE #646: BUG: Date parsing fails for non-UTC timezones
Describe the bug: Email date parsing may fail or show incorrect times for non-UTC timezones. No timezone normalization.

To reproduce:
1. Receive email with Date header in non-UTC timezone
2. Display shows wrong time or parse error
3. Sorting by date may be incorrect

Expected behavior: Parse all RFC 5322 timezone formats. Convert to local time for display. Handle parsing errors gracefully.

Additional context:
- File: fetcher/fetcher.go or date parsing code
- Use time.Parse() with RFC1123Z layout
- Fall back to multiple formats if first fails
- Severity: Medium - data display bug
- Complexity: Easy - proper time.Parse usage

andrinoff:
## What?

Converts date to local time, and enables highlighting the dates for when the email is selected for better visibility.

## Why?

Closes #646

### DIFF
--- a/main.go
+++ b/main.go
@@ -1127,7 +1127,7 @@ func (m *mainModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		if !strings.HasPrefix(normalizedSubject, "re:") {
 			subject = "Re: " + subject
 		}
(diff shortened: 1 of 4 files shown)


floatpanebot (approved):

PR #822, opened 2026-04-21
ci: fix screenshots workflow

andrinoff:
## What?

Fixes `theme_settings` because of updated settings in #683.

Also, updated the body of the PR to generate changed files.

## Why?

Previous workflow run was generating the wrong screenshot.


floatpanebot (approved):

PR #824, opened 2026-04-21
ci: add check for length

andrinoff:
## What?

Adds a check for the length of the title bar

## Why?

Squashed commits do not fit.


floatpanebot (approved):

PR #823, opened 2026-04-21
fix: validate port range

kyooosukedn:
## What?

- **`tui/login.go`**: New `validPort(s, fallback)` helper that parses and validates port strings against the valid TCP/UDP range (1–65535). Replaced raw `strconv.Atoi` calls for all three port fields (IMAP, SMTP, POP3). Invalid ports silently fall back to protocol defaults.
- **`tui/login_test.go`**: New test file with 24 test cases covering boundary values, negatives, overflow, non-numeric input, and end-to-end form submission.

## Why?

Fixes #740 — `tui/login.go` parsed port strings with `strconv.Atoi` but never checked if the value was in valid TCP range. Users could enter `0`, `-1`, or `70000` and get confusing connection errors with no indication of what went wrong.

PR #825, opened 2026-04-21
fix: limit filename length

### LINKED ISSUE #733: BUG: No validation on attachment filename length
**Describe the bug**
Attachment filename from email headers (untrusted) has no length validation. Extremely long filenames could cause filesystem errors or UI issues.

**To reproduce**
Related to #725 (path traversal). While that issue covers directory traversal, this covers length validation.

main.go:2603 uses origName directly without length check.

**Expected behavior**
Limit filename length to reasonable maximum (e.g., 255 bytes for most filesystems).

**Additional context**
- Good first issue
- Complements #725 sanitization
- Filesystem limits typically 255 bytes

**Suggested fix:**
```go
func sanitizeFilename(name string) string {
    name = filepath.Base(name)
    name = strings.ReplaceAll(name, "/", "_")
    name = strings.ReplaceAll(name, "\\", "_")
    
    // Limit length
    if len(name) > 200 {
        ext := filepath.Ext(name)
        name = name[:200-len(ext)] + ext
    }
    
    if name == "" || name == "." {
        name = "attachment"
    }
    
    return name
}
```

**OS**
All platforms

andrinoff:
## What?

Limited filename length to avoid system errors.

Refactored from #823, authored by @kyooosukedn 

## Why?

Fixes #733

### DIFF
--- a/main.go
+++ b/main.go
@@ -2584,6 +2584,16 @@ func sanitizeFilename(name string) string {
 	if name == "" || name == "." || strings.HasPrefix(name, ".") {
 		name = "attachment"
 	}
+	// Sanitize filename: enforce length limit to prevent filesystem errors
+	// with extremely long names from untrusted email headers.
+	const maxFilenameLen = 255
+	if len(name) > maxFilenameLen {
+		ext := filepath.Ext(name)
+		if len(ext) > maxFilenameLen {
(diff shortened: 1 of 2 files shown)


floatpanebot (approved):

PR #826, opened 2026-04-21
fix: load folders from cache

andrinoff:
## What?

Ensure INBOX is present, even if cache is empty or stale.

## Why?

Sometimes email folders will not load, causing a 2-5 second delay to switch folders


floatpanebot (dismissed):

floatpanebot (approved):

PR #827, opened 2026-04-21
fix: encryption settings not working

andrinoff:
## What?

Fixed a bug, where, if you go to encryption settings menu, you cannot enter the password, nor navigate. Also, fixed `esc` key handling.

## Why?

`updateEncryption` handles keys but doesn't forward input to textinput fields. Only navigation/enter handled, not typing.


floatpanebot (approved):

PR #832, opened 2026-04-21
fix: use crypto/rand for MIME boundary

### LINKED ISSUE #729: BUG: Weak random number generation for MIME boundaries
**Describe the bug**
In pgp/yubikey.go line 96, MIME boundary uses only time.Now().Unix() which is predictable and not cryptographically secure. While crypto/rand is used in sender.go, yubikey.go has weaker fallback.

**To reproduce**
Examine pgp/yubikey.go line 96:
```go
boundary := fmt.Sprintf("----=_Part_%d", time.Now().Unix())
```

Compare to sender/sender.go lines 81, 283, 487 which use crypto/rand properly.

**Expected behavior**
Use crypto/rand for MIME boundaries to prevent prediction attacks.

**Additional context**
- Good first issue
- Low severity (MIME boundary prediction has limited attack surface)
- Best practice: always use crypto/rand for boundaries
- sender/sender.go shows correct pattern

**Suggested fix:**
```go
buf := make([]byte, 16)
if _, err := rand.Read(buf); err == nil {
    boundary = fmt.Sprintf("----=_Part_%x", buf)
} else {
    // fallback
    boundary = fmt.Sprintf("----=_Part_%d", time.Now().UnixNano())
}
```

**OS**
All platforms

kyooosukedn:
## What?

Replaced `time.Now().Unix()` with `crypto/rand` for MIME boundary generation in `pgp/yubikey.go`.

## Why?

Fixes #729

The MIME boundary used a Unix timestamp, which is trivially predictable. While MIME boundary prediction has limited attack surface, `sender/sender.go` already uses `crypto/rand` for its boundaries. This brings `yubikey.go` in line with the established pattern.

Falls back to `time.Now().UnixNano()` if `crypto/rand` is unavailable (extremely unlikely).

## Testing

- `go build ./pgp/` — compiles clean

### DIFF
--- a/pgp/yubikey.go
+++ b/pgp/yubikey.go
@@ -3,6 +3,7 @@ package pgp
 import (
 	"bytes"
 	"crypto"
+	"crypto/rand"
 	"encoding/binary"
 	"fmt"
 	"io"
@@ -93,7 +94,7 @@ func BuildPGPSignedMessage(payload []byte, pin string, publicKeyPath string) ([]
 	headers, body := splitPayload(payload)
 
 	// Build the signed body part (this is what gets hashed)
-	boundary := fmt.Sprintf("----=_Part_%d", time.Now().Unix())
+	boundary := generateBoundary()
 	signedPart := buildSignedPart(headers, body, boundary)
 
 	// Build the OpenPGP signature packet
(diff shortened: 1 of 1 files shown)


floatpanebot (dismissed):
Hi @kyooosukedn! Please fix the following issues with your PR:

PR #833, opened 2026-04-21
fix: check tmpFile Close() error

### LINKED ISSUE #728: BUG: Unchecked tmpFile.Close() error in external editor
**Describe the bug**
In main.go line 2252, tmpFile.Close() is called but error is ignored. If Close() fails, writes may be lost but editor proceeds anyway.

**To reproduce**
Examine main.go lines 2245-2252:
```go
if _, err := tmpFile.WriteString(body); err != nil {
    tmpFile.Close()
    os.Remove(tmpPath)
    return func() tea.Msg {
        return tui.EditorFinishedMsg{Err: fmt.Errorf("writing temp file: %w", err)}
    }
}
tmpFile.Close()  // Error ignored
```

**Expected behavior**
Check Close() error before opening editor.

**Additional context**
- Good first issue
- Similar to #716
- Line 2246 Close() in error path also unchecked

**Suggested fix:**
```go
if _, err := tmpFile.WriteString(body); err != nil {
    tmpFile.Close()
    os.Remove(tmpPath)
    return func() tea.Msg {
        return tui.EditorFinishedMsg{Err: fmt.Errorf("writing temp file: %w", err)}
    }
}
if err := tmpFile.Close(); err != nil {
    os.Remove(tmpPath)
    return func() tea.Msg {
        return tui.EditorFinishedMsg{Err: fmt.Errorf("closing temp file: %w", err)}
    }
}
```

**OS**
All platforms

kyooosukedn:
## What?

Added error check on `tmpFile.Close()` in the external editor flow (`main.go`). Previously the error was ignored, so the editor could open with incomplete or empty content.

## Why?

Fixes #728

If `Close()` fails (e.g., disk full), data written via `WriteString()` may not be persisted to disk. The editor would then open an empty file, and the user could lose their draft content. Now returns an error to the UI instead of proceeding.

## Testing

- `go build .` — compiles clean

### DIFF
--- a/main.go
+++ b/main.go
@@ -2250,7 +2250,12 @@ func openExternalEditor(body string) tea.Cmd {
 			return tui.EditorFinishedMsg{Err: fmt.Errorf("writing temp file: %w", err)}
 		}
 	}
-	tmpFile.Close()
+	if err := tmpFile.Close(); err != nil {
+		os.Remove(tmpPath)
+		return func() tea.Msg {
+			return tui.EditorFinishedMsg{Err: fmt.Errorf("closing temp file: %w", err)}
+		}
+	}
 
 	parts := strings.Fields(editor)
 	args := append(parts[1:], tmpPath)


floatpanebot (dismissed):
Hi @kyooosukedn! Please fix the following issues with your PR:

- **Title**: Is too long (67 characters). The PR title must be strictly under 40 characters.

PR #842, opened 2026-04-22
feat: localization

andrinoff:
## What?

Adds support for several languages, with easy-to-manage JSON format.

Includes a new library `github.com/floatpane/matcha/i18n` a self-written library by us.

## Why?

Users would like support across their favorite languages.


floatpanebot (approved):

PR #843, opened 2026-04-22
fix: change the language dynamically

andrinoff:
## What?

Changes language dynamically (previously a reload was required.

## Why?

Users were seeing no results, when switching settings, and had no idea if it even works.


floatpanebot (approved):

PR #853, opened 2026-04-23
fix: remove html from linguist

andrinoff:
## What?

Removed `public/` from the linguist

## Why?

 
Misleading 3% HTML code


floatpanebot (approved):

PR #1015, opened 2026-04-25
ci: exclude all prereleases

andrinoff:
## What?

Excludes all prereleases from discord release webhook.

## Why?

PR previews were posted in the #release channel on discord, even though they didn't need to


floatpanebot (approved):

PR #1020, opened 2026-04-26
chore: use strings.Builder in validator

### LINKED ISSUE #963: BUG: String concatenation in loop in i18n validator report
### Describe the bug

In i18n/validator.go:127-139, report string is built with repeated `+=`:
```go
report += fmt.Sprintf("  [%s] %s: %s\n", err.Language, err.Key, err.Message)
report += fmt.Sprintf("  [%s] %d missing keys:\n", lang, len(keys))
```

For large validation reports with many languages and keys, this creates O(n²) allocations.

### To reproduce

Run i18n validation with many languages and missing keys.

### Expected behavior

Use `strings.Builder` instead of string concatenation.

mavonx:
## What?

Replaced repeated `+=` string concatenation in `i18n/validator.go` with `strings.Builder`.

## Why?

Each `+=` on a string allocates a new string, leading to `O(n²)` allocations when building large validation reports with many languages and missing keys.

Closes #963

### DIFF
--- a/i18n/validator.go
+++ b/i18n/validator.go
@@ -3,6 +3,7 @@ package i18n
 import (
 	"fmt"
 	"sort"
+	"strings"
 )
 
 // ValidationResult contains the results of validating translation files.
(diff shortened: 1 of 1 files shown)


andrinoff (approved):

andrinoff:
lgtm

PR #1021, opened 2026-04-26
ci: add /assign command

andrinoff:
## What?

Adds a `/assign` command for issues.

## Why?

So that users can assign issues to themselves, without having to ask maintainers.


floatpanebot (approved):

PR #1180, opened 2026-04-27
chore: defer file closes

### LINKED ISSUE #1056: BUG: Manual Close() instead of defer leaks fds on error in update copy
### Describe the bug

In main.go:3510-3525, the binary copy block opens two files and manually calls `Close` on every error path:
```go
in, err := os.Open(binPath)
if err != nil { ... }
out, err := os.OpenFile(tmpNew, ...)
if err != nil { in.Close(); ... }
if _, err := io.Copy(out, in); err != nil { in.Close(); out.Close(); ... }
in.Close()
out.Close()
```

This is fragile — every future return needs the same dance. A close error is also discarded, masking failed flushes.

### Expected behavior

Use `defer` for both files and report close errors (especially on `out`, where the close error is the only signal that pending data was actually flushed).

mavonx:
## What?

Replaced manual `Close()` calls on every error path in the binary copy block of `runUpdateCLI` with `defer`, and report close errors on `out`  via the named return value.

## Why?

The previous manual `Close()` calls on every return path were fragile and silently discarded flush errors on `out`.

Closes #1056

### DIFF
--- a/main.go
+++ b/main.go
@@ -3538,7 +3538,7 @@ func runUpdateCLI() (err error) {
 		}
 	}
 
-	if err := os.Rename(tmpNew, execPath); err != nil {
+	if err = os.Rename(tmpNew, execPath); err != nil {
 		return fmt.Errorf("could not replace executable: %w", err)
 	}
 
--- a/main.go
+++ b/main.go
@@ -3244,7 +3244,7 @@ func isFlagSet(fs *flag.FlagSet, name string) bool {
 	return found
 }
 
-func runUpdateCLI() error {
+func runUpdateCLI() (err error) {
 	const api = "https://api.github.com/repos/floatpane/matcha/releases/latest"
 	resp, err := httpClient.Get(api)
 	if err != nil {
@@ -3511,18 +3511,22 @@ func runUpdateCLI() error {
 	if err != nil {
 		return fmt.Errorf("could not open new binary: %w", err)
 	}
+	defer in.Close()
 	out, err := os.OpenFile(tmpNew, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0755)
 	if err != nil {
-		in.Close()
 		return fmt.Errorf("could not create temp binary in target dir: %w", err)
 	}
-	if _, err := io.Copy(out, in); err != nil {
-		in.Close()
-		out.Close()
+
+	defer func() {
+		cerr := out.Close()
+		if err == nil && cerr != nil {
+			err = fmt.Errorf("could not flush new binary to disk: %w", cerr)
+		}
+	}()
+
+	if _, err = io.Copy(out, in); err != nil {
 		return fmt.Errorf("could not write new binary to disk: %w", err)
 	}
-	in.Close()
-	out.Close()
 
 	// On Windows, a running executable cannot be overwritten directly.
 	// Move the old binary out of the way first, then rename the new one in.


Haroka-74:
@andrinoff

I need to change this `if err := os.Rename(tmpNew, execPath); err != nil {` to `if err = os.Rename(tmpNew, execPath); err != nil {`.

Am I right?

andrinoff:
> @andrinoff
> 
> I need to change this `if err := os.Rename(tmpNew, execPath); err != nil {` to `if err = os.Rename(tmpNew, execPath); err != nil {`.
> 
> Am I right?

I dont think so, although there could be a Windows-specific issue.
But it could be non-important and I couldn't get it to fail

Haroka-74:
Anyway, I'll change it for consistency.

andrinoff:
i'll check it out later (possibly tomorrow) @Haroka-74

main.go
```diff
@@ -3244,7 +3244,7 @@ func isFlagSet(fs *flag.FlagSet, name string) bool {
 	return found
 }
 
-func runUpdateCLI() error {
+func runUpdateCLI() (err error) {
```
andrinoff, line 3247: Not sure what is the reason for this implementation? this is heavily unnecessary

main.go
```diff
@@ -3511,18 +3511,22 @@ func runUpdateCLI() error {
...
-		out.Close()
+
+	defer func() {
+		cerr := out.Close()
+		if err == nil && cerr != nil {
+			err = fmt.Errorf("could not flush new binary to disk: %w", cerr)
+		}
+	}()
+
+	if _, err = io.Copy(out, in); err != nil {
```
andrinoff, line 3527: if removed from the above, this has to be a declaration (`:=`)

andrinoff (changes requested):
@Haroka-74 fix the comments

Haroka-74:
Hey @andrinoff, just to clarify why the named return is necessary.

The original issue asked for two things:
- Use `defer` for both files
- Report close errors (especially on `out`, where a close error is the only signal that pending data was actually flushed)

To satisfy point `2`, the defer needs a way to propagate the close error back to the caller. The only way to do that from a defer is via a named return:

```go
func runUpdateCLI() (err error) {
    // ...
    defer func() {
        if cerr := out.Close(); cerr != nil && err == nil {
            err = fmt.Errorf("could not flush new binary to disk: %w", cerr)
        }
    }()
    // ...
}
```

Without the named return, the `defer` has no way to surface the close error to the caller — it would just be silently dropped, which is exactly what the issue was complaining about.

Let me know what you think, happy to discuss if you have a different approach in mind!

andrinoff:
An IIFE removes the need for a named return entirely. Each file gets its own defer scoped to the closure, and errors are returned through a normal local variable:

example:

```go
var copyErr error
func() {
    defer in.Close()
    defer func() {
        if cerr := out.Close(); copyErr == nil && cerr != nil {
            copyErr = fmt.Errorf("could not flush new binary to disk: %w", cerr)
        }
    }()
    if _, copyErr = io.Copy(out, in); copyErr != nil {
        copyErr = fmt.Errorf("could not write new binary to disk: %w", copyErr)
    }
}()
if copyErr != nil {
    return copyErr
}
```

andrinoff (approved):
lgtm

andrinoff:
i'll merge the PR, it is good enough and pretty clean, though i do not like the implementation entirely

floatpanebot (approved):

Haroka-74:
I can change the implementation if you’d like.

andrinoff:
> I can change the implementation if you’d like.

already merged and releasing, if the issue comes up, i'll link this PR

PR #1184, opened 2026-04-28
fix: handle zero terminal cell height

### LINKED ISSUE #865: BUG: Division by zero potential in imageRows when cellHeight is 0
### Describe the bug

In `imageRows()` (view/html.go:274-278), `cellHeight` from `getCellHeightFromFd()` could return 0 if the terminal doesn't support the ioctl query. The division `(h + cellHeight - 1) / cellHeight` would then panic with division by zero.

### To reproduce

1. Run matcha in a terminal that doesn't report cell size via ioctl
2. Open an email with an inline image
3. Observe division by zero panic

### Expected behavior

Check `cellHeight > 0` before dividing, default to a reasonable fallback (e.g. 16px).

mavonx:
## What?

Added a check for `cellHeight == 0` in the `imageRows` function within `view/html.go`. If the terminal height cannot be determined, it now defaults to a standard `16px` fallback.

## Why?

When running in terminals that don't report cell size via `ioctl`, `getTerminalCellSize()` returns `0`. This previously caused a "division by zero" runtime panic during the row calculation for inline images: `(h + cellHeight - 1) / cellHeight`.

Closes #865

### DIFF
--- a/view/html.go
+++ b/view/html.go
@@ -509,6 +509,9 @@ func imageRows(payload string) int {
 	if data, err := base64.StdEncoding.DecodeString(payload); err == nil {
 		if _, h, ok := clib.ImageDimensions(data); ok {
 			cellHeight := getTerminalCellSize()
+			if cellHeight == 0 {
+				cellHeight = 16
+			}
 			rows = (h + cellHeight - 1) / cellHeight
 			if rows < 1 {
 				rows = 1


floatpanebot (dismissed):

- **Title**: Is too long (56 characters). The PR title must be strictly under 40 characters.

floatpanebot (dismissed):

- **Title**: Is too long (43 characters). The PR title must be strictly under 40 characters.

LeaWhoCodes (dismissed):
lgtm

LeaWhoCodes:
@Haroka-74, actually, found a piece of code 

```go
func getTerminalCellSize() int {
	const defaultCellHeight = 18

	// Try stdout, stdin, stderr, then /dev/tty as last resort
	fds := []int{int(os.Stdout.Fd()), int(os.Stdin.Fd()), int(os.Stderr.Fd())}

	for _, fd := range fds {
		if cellHeight := getCellHeightFromFd(fd); cellHeight > 0 {
			return cellHeight
		}
	}

	// Try /dev/tty directly - this works even when stdio is redirected (e.g., in Bubble Tea)
	if tty, err := os.Open("/dev/tty"); err == nil {
		defer tty.Close()
		if cellHeight := getCellHeightFromFd(int(tty.Fd())); cellHeight > 0 {
			return cellHeight
		}
	}

	debugImageProtocol("using default cell height: %d pixels", defaultCellHeight)
	return defaultCellHeight
}
```
`html.go:28-50`

where we do have a fallback value. @andrinoff will have to check this out later, he is extremely busy right now, so no promises

Haroka-74:
@LeaWhoCodes 

Good point, I didn’t notice `const defaultCellHeight = 18`.

LeaWhoCodes:
> @LeaWhoCodes
> 
> Good point, I didn’t notice `const defaultCellHeight = 18`.

this might still be a good fix, if it is determined that we can find a `0` in the function

andrinoff (approved):
Good fallback in case the system says the `cellHeight = 0`.

lgtm

floatpanebot (approved):

Issue #1131, opened 2026-04-27
FEAT: Server-side IMAP SEARCH support


andrinoff:
### Describe the feature

Plumb the IMAP `SEARCH` command (and `ESEARCH` extension when supported) so users can search the entire mailbox without paginating through every UID. Pair it with the local index — local hits show instantly; the server search backfills.

### Why this matters

For inboxes with 100k+ messages, downloading everything to build a local index is impractical at first run. Server-side SEARCH gives an instant first-pass result while the local index warms up.

### Notes

- Map our search DSL to IMAP search keys: `from:`, `to:`, `subject:`, `body:`, `since:`, `larger:`.
- Use `ESEARCH RETURN (ALL)` (RFC 4731) to get UID lists in one round-trip.
- Gracefully degrade to `SEARCH UID` for old servers.

floatpanebot:
This has had no activity for 30 days. It will be closed in 14 days unless updated. Comment or remove the `stale` label to keep it open.

floatpanebot:
Closing due to inactivity. Reopen if still relevant.

Issue #1129, opened 2026-04-27
FEAT: Full-text search index across folders and accounts


andrinoff:
### Describe the feature

A first-class email client lets you find any message in seconds. Right now matcha can only filter what's loaded into the current folder view.

Ship a local full-text search index (SQLite FTS5 or Bleve) that:
- ingests every fetched message body + headers
- indexes per-account, per-folder
- supports prefix, phrase, and field-qualified queries (`from:alice subject:invoice`)
- updates incrementally on IDLE notifications
- exposes a TUI search modal (`/`) and a `matcha search` CLI command

### Why this matters

Search is table stakes for any mail client; Mutt, aerc, neomutt, Thunderbird all have it. Without it the user has to fall back to grepping raw IMAP downloads.

### Notes

- SQLite FTS5 is probably the cheapest path: zero deps, atomic writes, ~tens of MB for a heavy inbox.
- Index should live in `~/.cache/matcha/search.db` so it can be rebuilt without losing config.
- Tokenizer needs to handle CJK / Cyrillic / accented Latin reasonably (FTS5 unicode61 tokenizer).

floatpanebot:
This has had no activity for 30 days. It will be closed in 14 days unless updated. Comment or remove the `stale` label to keep it open.

floatpanebot:
Closing due to inactivity. Reopen if still relevant.

PR #1182, opened 2026-04-28
fix(sender): check writes in iMIP body

mvanhorn:
## What?

`SendCalendarReply` in `sender/sender.go` built its multipart iMIP body via direct `Write` / `Fprint` / `Close` calls whose errors were discarded. This PR wraps each of the eight unchecked calls in the iMIP body block in the standard Go `if _, err := ...; err != nil { return nil, err }` idiom.

The eight callsites covered:

- `fmt.Fprint(qp, plainBody)`
- `qp.Close()`
- `calPart.Write(...)`
- `altWriter.Close()`
- `altPart.Write(...)`
- `attachPart.Write(...)`
- `outerWriter.Close()`
- `msg.Write(...)`

## Why?

A short write or buffer error in any of these calls produces a malformed iMIP reply. Google Calendar and Outlook silently reject malformed replies without recording the RSVP, so the user thinks they accepted the meeting but the organizer never sees it. Capturing the error and returning early makes the failure mode loud (caller gets an error) instead of silent (corrupt outbound mail).

No behavior change on the success path.

Closes #1128.

PR #1027, opened 2026-04-27
chore(daemon): named timeout constants

mvanhorn:
## What?

Replaces ad-hoc `60*time.Second` and `30*time.Second` literals in `daemon/handler.go` with two named constants and a brief comment documenting the split, exactly as suggested in the issue body.

```go
const (
    fetchTimeout  = 60 * time.Second
    mutateTimeout = 30 * time.Second
)
```

`fetchTimeout` (was `60*time.Second`):
- `handleFetchEmails` (line 129)
- `handleFetchEmailBody` (line 154)
- async fetch inside `handleMarkRead` (line 318)

`mutateTimeout` (was `30*time.Second`):
- `handleDeleteEmails` (line 193)
- `handleArchiveEmails` (line 216)
- `handleMoveEmails` (line 239)
- `handleMarkRead` (line 262)
- `handleFetchFolders` (line 287)

## Why?

> `fetchTimeout` covers reads against the upstream IMAP provider, which can return large bodies and so are given more headroom. `mutateTimeout` covers state-changing operations and folder listings, which are bounded by IMAP command latency rather than payload size.

If the maintainer wants `handleFetchFolders` reclassified to `fetchTimeout` (which would be a behavior change from 30s to 60s), happy to do that as a follow-up — kept it at the existing value here to keep this purely a refactor.

Closes #985.

PR #1186, opened 2026-04-28
feat(search): server-side email search

### LINKED ISSUE #508: FEAT: Email search functionality

There is no way to search emails in matcha. Users with large mailboxes have no option other than scrolling to find specific emails. Search is a fundamental email client feature.

1. Add a `Search(ctx context.Context, query string, folder string) ([]Email, error)` method to the `backend.Provider` interface
2. Implement using IMAP `SEARCH` command for the IMAP backend
3. Implement using JMAP `Email/query` filters for the JMAP backend
4. Add a search TUI overlay (e.g. triggered by `/` key) with a text input and results list
5. Support searching by sender, subject, body text, and date range

Client-side filtering of already-fetched emails. This would be simpler but only works for cached emails and doesn't scale to large mailboxes.

### Additional context

Both IMAP and JMAP protocols have native server-side search capabilities that are already available in the protocol libraries used by matcha (`go-imap/v2` and `go-jmap`) but are not currently exposed.

mvanhorn:
## What?

- Add `EmailSearcher` interface (`Search(ctx, folder, SearchQuery) ([]Email, error)`) to `backend.Provider`
- Add `SearchQuery` struct + `ParseSearchQuery` DSL parser supporting `from:`, `to:`, `subject:`, `body:`, `since:`, `before:`, `larger:`, plus quoted multi-word values and bare body terms
- IMAP backend: `fetcher.SearchMailbox` uses `go-imap/v2` `UIDSearch`, prefers `ESEARCH RETURN (ALL)` (RFC 4731) when `imap.CapESearch` or `imap.CapIMAP4rev2` is advertised, falls back to UID `SEARCH` for older servers
- JMAP backend: `Email/query` with `FilterCondition` (`From`, `To`, `Subject`, `Body`, `After`, `Before`, `MinSize`)
- POP3 backend: returns `backend.ErrNotSupported`
- TUI: new `SearchOverlay` triggered by `/` from inbox, results render below the email list, Enter applies them as a temporary "Search Results" view, Esc clears
- Multi-account: `/` from "All Accounts" runs across every supported account; `ErrNotSupported` is skipped silently so a mixed IMAP+JMAP+POP3 setup still surfaces results from the supported backends
- Default keybind `inbox.search` = `/` in `default_keybinds.json`
- Demo helper at `screenshots/cmd/search_view/main.go` + `screenshots/search_demo.tape` so the demo regenerates without IMAP credentials
- Tests: parser DSL (quoted values, body precedence, mixed prefixed+bare), JMAP filter assembly, fetcher criteria mapping, POP3 unsupported

## Why?

Matches the spec from issue #508 directly:

> Add a `Search(ctx context.Context, query string, folder string) ([]Email, error)` method to the `backend.Provider` interface ... Implement using IMAP `SEARCH` ... JMAP `Email/query` filters ... search TUI overlay (e.g. triggered by `/` key)

Sibling issue #1131 added the ESEARCH + DSL prefix detail (`from:`, `to:`, `subject:`, `body:`, `since:`, `larger:`, ESEARCH RETURN ALL with UID SEARCH fallback).

The launch thread on r/coolgithubprojects had filtering as the very first comment ("By chance are there filtering options?"), so server-side search closes both the maintainer's and a user-visible gap.

Closes #508. Partially addresses #1131. Out of scope: #1129 (full-text index across folders).


floatpanebot (dismissed):

- **Title**: Is too long (62 characters). The PR title must be strictly under 40 characters.

andrinoff:
Okay, this is a great issue to implement!

There are a few problems, right off the bat, not looking at code:

1. Emails are shown for `email` property, it is not shown by `fetch_email`. Issue: some accounts (like mine, personally) have separate `fetch_email`'s and I have for all of them shown the same email, even though it is addressed to only 1 of them. 
2. The account filtering selector does not work in the search mode
3. You removed the quick search, not sure if that is a good decision. The fuzzy search we had before allowed you to search across cached emails, which was pretty useful, I'd leave both of these.
4. run `make lint` (the reason CI is failing)
5. Finally, the emails that are not loaded cannot be opened (you need to fetch the UUID's if you don't already, I haven't checked out the code yet)

Please, fix these things first, and I'll get around checking this PR and refining if needed later (may take several days, quite busy atm)

mvanhorn:
Pushed cff22ef addressing all three points:

1. Search results and the All-Accounts view now resolve the per-row label to the To: recipient that matches the account's FetchEmail (parsed via net/mail), so when a single upstream mailbox serves multiple FetchEmails the row shows which one it was actually addressed to. Falls back to the account's FetchEmail when no recipient matches. Also fixed a smaller bug in SetEmails which was using acc.Email instead of FetchEmail for tab labels.
2. Switching the account tab while search is active now narrows searchResults in-memory (new filteredSearchResults helper), so the cross-account search you run from All-Accounts can be filtered to one account without re-running the query.
3. Restored the client-side bubble-list filter under a new inbox.filter keybind (default f). / still launches the server-side search overlay. The two are complementary: f for instant in-memory filtering of what's loaded, / for going back to the server.

Verified: go test ./tui/... passes locally. Also fixed the lint failure on screenshots/cmd/search_view/main.go. Let me know if you want any of these tuned differently, particularly the keybind choice for filter.

andrinoff:
> 1. Search results and the All-Accounts view now resolve the per-row label to the To: recipient that matches the account's FetchEmail (parsed via net/mail), so when a single upstream mailbox serves multiple FetchEmails the row shows which one it was actually addressed to. Falls back to the account's FetchEmail when no recipient matches. Also fixed a smaller bug in SetEmails which was using acc.Email instead of FetchEmail for tab labels.

this still doesnt work

image


as well as that, still cannot open emails that are not loaded in the cache.

mvanhorn:
Pushed 870aa1d addressing the two remaining points.

1. The All-Accounts view and search results now dedupe by Message-ID (with a `From|Subject|Date.UnixNano()` fallback for malformed messages). When one upstream mailbox serves multiple FetchEmails the same message shows up once instead of once per account; the surviving row prefers the copy whose owning account's FetchEmail matches a `To:` recipient, and `accountLabelForEmail` resolves across the full accounts list instead of just the fetching account. With your `edu`/`me`/`business@andrinoff.com` setup the message in your screenshot should now appear once, labeled `business@andrinoff.com`. Per-account tab views are unchanged.

5. `ViewEmailMsg` now carries an optional `*fetcher.Email`. The Inbox embeds it when opening a row from search results, so opening a search hit that wasn't in the local cache no longer silently drops the keypress. `main.go` prefers `msg.Email`, adds the email to `m.emailsByAcct`/`m.emails` when missing (so next/prev navigation still works), and falls through to the existing `FetchEmailBody` path.

Verified locally: `go test ./...` green, `gofmt -l .` clean. New tests cover the dedup and uncached-open paths: `TestInboxAllAccountsDedupesSharedMailboxByMessageID`, `TestInboxSearchResultsDedupedAcrossAccounts`, `TestInboxAllAccountsDoesNotDedupeWhenMessageIDDiffers`, `TestInboxOpenSearchResultEmbedsEmailInViewMsg`. `TestInboxAccountLabelUsesMatchingRecipient` updated to assert the new cross-account match.

LeaWhoCodes:
> 2\. `ViewEmailMsg` now carries an optional `*fetcher.Email`. The Inbox embeds it when opening a row from search results, so opening a search hit that wasn't in the local cache no longer silently drops the keypress. `main.go` prefers `msg.Email`, adds the email to `m.emailsByAcct`/`m.emails` when missing (so next/prev navigation still works), and falls through to the existing `FetchEmailBody` path.

interesting, haven't looked at the diff, but the issue seems to persist with split view mode (#1006), although, it does work in full screen mode, maybe you can find the culprit

mvanhorn:
Pushed 8fc9e1d. Root cause was a parallel split-pane code path: OpenSplitPreview discarded ViewEmailMsg.Email and PreviewBodyFetchedMsg.findEmailByUID only searched m.inbox.allEmails, which never contains search hits. Threaded the resolved email through OpenSplitPreview onto FolderInbox.previewSearchEmail and made findEmailByUID fall back to it when allEmails has no match. Verified locally: go test ./... is green, including two new tests in tui/folder_inbox_test.go that cover the cross-folder search-hit case and confirm the live allEmails entry still wins when both are present.

andrinoff (dismissed):
lgtm after fixing

andrinoff (approved):
lgtm

floatpanebot (approved):

mvanhorn:
Thanks for the merge run this week (#1007, #1027, #1182, #1190, and now this one). Glad the ESEARCH-with-fallback path landed for older IMAP servers.

Issue #514, opened 2026-04-15
FEAT: Plugin API - multi-field prompts and confirm dialogs


andrinoff:
### Is your feature request related to a problem?

`matcha.prompt()` only supports a single text input field. Plugins that need multi-step configuration (e.g. API key + model selection, or multiple form fields) must chain multiple prompts sequentially, which is awkward UX. There's also no way to ask a simple yes/no confirmation question.

### Describe the solution you'd like

At minimum, add a confirm dialog:

```lua
matcha.confirm("Are you sure you want to proceed?", function(yes)
    if yes then
        -- do the thing
    end
end)
```

Ideally, also support multi-field prompts:

```lua
matcha.prompt_fields({
    {name = "api_key", placeholder = "Enter API key"},
    {name = "model", placeholder = "Model name (default: gpt-4)"},
}, function(fields)
    local key = fields.api_key
    local model = fields.model
end)
```

### Describe alternatives you've considered

Chaining multiple `matcha.prompt()` calls in nested callbacks. This works but creates deeply nested callback hell and poor UX (multiple sequential popups).

### Additional context

N/A

floatpanebot:
This has had no activity for 30 days. It will be closed in 14 days unless updated. Comment or remove the `stale` label to keep it open.

floatpanebot:
Closing due to inactivity. Reopen if still relevant.

Issue #513, opened 2026-04-15
FEAT: Plugin API - account info API


andrinoff:
### Is your feature request related to a problem?

Plugins have no way to query account information. There is no `matcha.accounts()` or `matcha.current_account()` API. Multi-account setups can't have account-specific plugin behavior (e.g. different signatures per account, different auto-BCC rules, account-specific workflows).

### Describe the solution you'd like

Add API functions:

```lua
-- Get the currently active account
local acct = matcha.current_account()
-- acct.email, acct.name, acct.id

-- List all configured accounts
local accounts = matcha.accounts()
for _, acct in ipairs(accounts) do
    matcha.log(acct.email)
end
```

Only expose safe, read-only fields (email, display name, account ID). Do not expose passwords, tokens, or server credentials.

### Describe alternatives you've considered

Using the `account_id` field already present in hook email tables and hardcoding known IDs. This is fragile and doesn't let plugins discover accounts dynamically.

### Additional context

N/A

floatpanebot:
This has had no activity for 30 days. It will be closed in 14 days unless updated. Comment or remove the `stale` label to keep it open.

floatpanebot:
Closing due to inactivity. Reopen if still relevant.

Issue #512, opened 2026-04-15
FEAT: Plugin API - missing lifecycle hooks (delete/archive/move)


andrinoff:
### Is your feature request related to a problem?

The plugin hook system only has 8 events: `startup`, `shutdown`, `email_received`, `email_viewed`, `email_send_before`, `email_send_after`, `folder_changed`, `composer_updated`. There are no hooks for destructive/organizational email actions, so plugins can't react to email lifecycle changes.

### Describe the solution you'd like

Add the following hooks to `plugin/hooks.go`:

- `email_deleted` — called when an email is deleted (with uid, account_id, folder)
- `email_archived` — called when an email is archived
- `email_moved` — called when an email is moved (with source and destination folder)
- `attachment_downloaded` — called when an attachment is saved

### Describe alternatives you've considered

Polling-based approach where plugins periodically check folder contents. This is inefficient and misses the specific action taken.

### Additional context

These hooks would enable plugins like: deletion audit logs, sync with external task managers (e.g. mark task complete when email archived), analytics dashboards, and backup plugins.

floatpanebot:
This has had no activity for 30 days. It will be closed in 14 days unless updated. Comment or remove the `stale` label to keep it open.

floatpanebot:
Closing due to inactivity. Reopen if still relevant.

Issue #511, opened 2026-04-15
FEAT: Plugin API - email body access in hooks


andrinoff:
### Is your feature request related to a problem?

The `email_viewed` hook receives email metadata (uid, from, to, subject, date, is_read) but NOT the email body. Plugins that want to analyze, summarize, or act on email content cannot do so. This limits plugins to metadata-only operations.

### Describe the solution you'd like

Either:
1. Include the email body (plain text) in the `email_viewed` hook table, or
2. Add a `matcha.get_email_body(uid, account_id)` function that fetches the body on demand

Option 2 is preferred as it avoids loading large bodies when plugins don't need them.

### Describe alternatives you've considered

Passing the body in the hook table directly. This is simpler but wastes memory/CPU when plugins only care about metadata.

### Additional context

This would enable plugins like: email summarizers, link extractors, auto-labelers, content-based notifications, and AI-powered email analysis.

floatpanebot:
This has had no activity for 30 days. It will be closed in 14 days unless updated. Comment or remove the `stale` label to keep it open.

floatpanebot:
Closing due to inactivity. Reopen if still relevant.

PR #1187, opened 2026-04-28
feat(plugin): per-plugin KV storage

### LINKED ISSUE #510: FEAT: Plugin API - persistent storage

Plugins have no way to persist data between sessions. Plugins like the AI rewrite plugin can't save API keys, user preferences, or cached data without relying on external HTTP calls. Each session starts with a blank slate.

Add a key-value storage API scoped per plugin:

```lua
local matcha = require("matcha")

-- Store a value
matcha.store_set("api_key", "sk-...")

-- Retrieve a value
local key = matcha.store_get("api_key")
```

Backend: a JSON file per plugin at `~/.config/matcha/plugins/<name>/data.json`. Values should be strings (plugins can JSON-encode complex data themselves).

- Letting plugins use `io` library for file access — but this breaks the sandbox security model
- Environment variables — not persistent and pollutes the shell environment
- Using `matcha.http()` to store data externally — requires network and an external service

### Additional context

This is the most-requested missing API for plugin authors. It would unblock use cases like: saving API keys, caching results, storing per-user preferences, tracking plugin state.

mvanhorn:
## What?

- Add `matcha.store_set(key, value)`, `store_get(key)`, `store_delete(key)`, `store_keys()` to the Lua plugin API
- Storage lives at `~/.config/matcha/plugins/<plugin_name>/data.json` with mode `0o600`, JSON-encoded, atomic write via unique temp file + `Rename`
- Per-plugin scoping: Manager tracks the active plugin during plugin load, hook invocation, and keybinding callbacks; hooks/keybindings now carry their plugin attribution (`registeredHook{fn, plugin}`)
- Plugin-name validation: only `^[a-zA-Z0-9_-]+$` is accepted as a path component, blocking traversal
- Lua API surfaces real store-init errors (corrupt JSON, permission denied) to plugin authors instead of swallowing them as "no plugin context"
- Tests cover: round-trip set/get/delete/keys, persistence across `Manager` instances, concurrent writes, file mode `0o600` (including survives overwrite), invalid plugin name rejection, Lua-level error propagation on corrupt JSON, hook/keybinding plugin attribution
- Demo helper: `screenshots/cmd/plugin_storage_demo/main.go` loads the same plugin in two `Manager` instances against the same `HOME` to prove cross-session persistence

demo

*Simulated demo (Remotion) — matcha theme, scripted UI, not a live capture.*

## Why?

This is the maintainer-spec from issue #510:

> Plugins have no way to persist data between sessions. Plugins like the AI rewrite plugin can't save API keys, user preferences, or cached data without relying on external HTTP calls. Each session starts with a blank slate.

The four function names, JSON-per-plugin file format, and config-dir path all match the spec verbatim:

```lua
local matcha = require("matcha")
matcha.store_set("api_key", "sk-...")
local key = matcha.store_get("api_key")
```

#510 is the gating issue for the broader plugin-API cluster (#511 hooks, #512 lifecycle, #513 account info, #514 multi-field prompts) - persistent storage is the most-requested missing piece for the existing 35-plugin marketplace.

## Notes

The implementation extends the per-plugin attribution model so future plugin APIs (account info, lifecycle hooks) can reuse the same `currentPlugin` plumbing in `Manager`.

Closes #510.

This contribution was developed with AI assistance.


mvanhorn:
Pushed 138272f. The Windows failures came down to t.Setenv("HOME", t.TempDir()) not isolating anything on Windows because os.UserHomeDir() reads %USERPROFILE%, not $HOME, so every test in the package shared the real Windows profile and bled state. Added a setTestHome helper that sets both, and guarded the 0o600 mode assertions on Windows since NTFS does not honor Unix permissions the same way. go test ./plugin/... passes locally on macOS.

andrinoff:
@mvanhorn update documentation at `docs/docs/Features/Plugins.md` and `plugin/README.md`

mvanhorn:
Pushed bd3079b - added the matcha.store_set / store_get / store_delete / store_keys section to docs/docs/Features/Plugins.md (after matcha.notify, before Events). plugin/README.md already had the storage API + Persistent storage section in this PR.

andrinoff (approved):

mvanhorn:
Thanks for landing the per-plugin KV storage, @andrinoff. Took a couple rounds on the Windows test isolation, but it's clean now.

PR #1190, opened 2026-04-29
fix(cache): trim <> from email keys

### LINKED ISSUE #1122: BUG: contact normalization does not strip angle brackets
### Describe the bug

`config/cache.go:151-153`:

```go
func normalizeContactEmail(email string) string {
    return strings.ToLower(strings.Trim(strings.TrimSpace(email), ","))
}
```

Email addresses passed in from `From:` headers commonly arrive as `<foo@bar.com>`. The angle brackets are not stripped, so the cached entry stores `<foo@bar.com>` literally and never matches future lookups against `foo@bar.com`.

### Expected behavior

Strip `<` and `>` (and any leading/trailing whitespace) before lowercasing.

mvanhorn:
## What?

Add `<` and `>` to the trim cutset in `normalizeContactEmail` so the cache key is the bare email regardless of whether the input came from a header or a UI field. Whitespace and commas continue to be trimmed.

## Why?

Email addresses arriving from `From:` headers commonly come wrapped in angle brackets (`<foo@bar.com>`). The previous trim cutset was `,` only, so the cache stored the literal `<foo@bar.com>` and never matched future lookups against `foo@bar.com`.

Resolves #1122.

### DIFF
--- a/config/cache.go
+++ b/config/cache.go
@@ -149,7 +149,7 @@ func LoadContactsCache() (*ContactsCache, error) {
 }
 
 func normalizeContactEmail(email string) string {
-	return strings.ToLower(strings.Trim(strings.TrimSpace(email), ","))
+	return strings.ToLower(strings.Trim(strings.TrimSpace(email), ",<>"))
 }
 
 // AddContact adds or updates a contact in the cache.


floatpanebot (dismissed):

- **Title**: Is too long (57 characters). The PR title must be strictly under 40 characters.

mvanhorn:
Updated title to `fix(cache): trim <> from email keys` (35 chars) and rewrote the body with `## What?` / `## Why?` headings. No code changes.

andrinoff (approved):
lgtm

floatpanebot (approved):

PR #1194, opened 2026-04-29
fix: notifications being sent when off

andrinoff:
## What?

Added a reload config handler for the daemon and daemon reload.

## Why?

There was an issue with the daemon sending notifications, even if config does not have the option turned on


floatpanebot (approved):

PR #1198, opened 2026-04-29
fix(i18n): wrap original error in i18n

mavonx:
## What?

Replaced the `%v` verb with `%w` in `i18n/parser.go` when returning errors from the parser.

## Why?

The previous implementation only wrapped `ErrParseFailed`. By rendering the underlying err as a string (`%v`), the error chain was broken. This prevented callers from using `errors.Is()` or `errors.As()` to detect the original cause of the failure. Using double `%w` (supported in Go 1.20+) ensures both the sentinel error and the root cause remain accessible in the error tree.


floatpanebot (dismissed):

- **Title**: Is too long (47 characters). The PR title must be strictly under 40 characters.

PR #1202, opened 2026-04-30
fix(i18n): wrap error in loader

### LINKED ISSUE #1051: BUG: fmt.Errorf chain '%w: %v' loses original cause in embedded loader
### Describe the bug

In i18n/loader.go:15:
```go
return fmt.Errorf("%w: embedded load failed: %v", ErrLoadFailed, err)
```

Same pattern as the parser: the underlying load error is stringified and dropped from the chain.

### Expected behavior

Wrap both errors with `%w`.

mavonx:
## What?

Replaced the `%v` verb with `%w` in `i18n/loader.go` when returning errors from the embedded loader.

## Why?

The current implementation stringifies the underlying `err` using `%v`, which breaks the error chain. This prevents callers from inspecting the root cause using `errors.Is()` or `errors.As()`. Utilizing double `%w` (supported in Go 1.20+) ensures both `ErrLoadFailed` and the original error remain accessible in the error tree.

Closes #1051

### DIFF
--- a/i18n/loader.go
+++ b/i18n/loader.go
@@ -13,7 +13,7 @@ import (
 func LoadTranslations(bundle *Bundle) error {
 	// Load from embedded files
 	if err := loadFromEmbedded(bundle); err != nil {
-		return fmt.Errorf("%w: embedded load failed: %v", ErrLoadFailed, err)
+		return fmt.Errorf("%w: embedded load failed: %w", ErrLoadFailed, err)
 	}
 
 	return nil


andrinoff (approved):
lgtm, but @Haroka-74 , please pick up non good first issues too, we try to leave the good first issues for newcomers

Haroka-74:
Totally understand.

I’m still relatively new to the Go ecosystem, so I started with some good first issues to get familiar with the codebase and workflow. I’ll focus more on non good first issues going forward.

floatpanebot (approved):

PR #1203, opened 2026-04-30
fix: pass through in search

### LINKED ISSUE #1199: BUG: `m` key triggers `move` in search field
### Describe the bug

Thanks for this excellent commit -- and Matcha altogether: https://github.com/floatpane/matcha/commit/72e6d3ca493ed7b5527910065c6001737481613d

When I enter search, the `m` key triggers the `move` command instead of entering text.

All other keys from `keybindings.json` (\*) work as expected and just enter text in the search field.

\* `tab` triggers `next_folder`. But that's expected and handy.

### To reproduce

- `Enter` to "View Inbox"
- `/`
- `m` ("Move to folder:" popup appears)

### Expected behavior

`m` in the search field should print the letter "m".

### Screenshots

_No response_

### Matcha version

v0.36.0-2-gba6b659

### Go version

1.26.2

### OS

Void Linux

### Terminal

foot

### Additional context

_No response_

mvanhorn:
## What?

In `tui/folder_inbox.go`, skip the `FolderInbox.Update` keypress switch while `m.inbox.searchOverlay != nil` so the message falls through to `m.inbox.Update`, which already routes overlay keys to `SearchOverlay.Update` for text input.

Adds `TestSearchOverlayKeysNotIntercepted` in `tui/folder_inbox_test.go` to lock in the behavior. The test fails on `master` and passes after the fix.

## Why?

Closes #1199.

`FolderInbox.Update` runs before the inner `Inbox.Update`. When the search overlay was open (after pressing `/`), folder-level keybindings such as `kb.Folder.Move` (default `m`) matched on `msg.String()` and started the move-to-folder flow instead of letting the keystroke reach the search input. Reproducer from the issue: enter inbox -> `/` -> press `m` and the "Move to folder:" popup appears.

Other `kb.Folder.*` bindings (`tab`, `shift+tab`, `]`, `[`) had the same problem in principle. Routing all keys through to the inner inbox while the overlay is open fixes them in one place. The overlay's own `Esc` handler in `Inbox.Update` (line 647-650) still works because the inbox's keypress branch checks `m.searchOverlay != nil` first.

### DIFF
--- a/tui/folder_inbox.go
+++ b/tui/folder_inbox.go
@@ -172,6 +172,12 @@ func (m *FolderInbox) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			break
 		}
 
+		// Don't intercept keys while the inbox search overlay is active.
+		// Otherwise folder-level bindings like "m" (move) would shadow text input.
+		if m.inbox.searchOverlay != nil {
+			break
+		}
+
 		kb := config.Keybinds
 
 		// Route input to preview pane when focused
--- a/tui/folder_inbox_test.go
+++ b/tui/folder_inbox_test.go
@@ -86,3 +86,39 @@ func TestFolderInboxSplitPreviewPrefersAllEmails(t *testing.T) {
 		t.Fatalf("expected the live allEmails entry, got %+v", got)
 	}
 }
+
+// TestSearchOverlayKeysNotIntercepted covers issue #1199: pressing keys that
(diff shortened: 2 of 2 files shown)


floatpanebot (dismissed):

- **Title**: Is too long (52 characters). The PR title must be strictly under 40 characters.

andrinoff (approved):
lgtm

floatpanebot (approved):

PR #1206, opened 2026-04-30
fix: allow subaddresses for gmail

### LINKED ISSUE #1200: FEAT: Gmail dots-don't-matter and subaddressing

[In Gmail, dots don't matter](https://support.google.com/mail/answer/7436150). [Subaddressing](https://gmail.googleblog.com/2008/03/2-hidden-ways-to-get-more-from-your.html) is also relatively widely known, and I assume used.
When adding a new email account, you have to supply a comma-separated list of addresses for which to fetch mail for. That is infeasible when there can be infinite addresses with these two things.

Right now, if you connect your gmail account, any mail that isn't addressed to an explicitly set fetch address doesn't show up, and since I make heavy use of subaddressing, Matcha is basically unusable for me with Gmail ☹️

I can see a couple of ways to approach it. Either you could support regex or globbing for fetch addresses. Advantages are that its easily extensible, and can apply to more than just Gmail. Disadvantages are that its a bit complex and still won't solve the problem without a default matcher that the user is informed about.

The other way would be to hard code these rules into the gmail provider. Not extensible, but simple.

_No response_

### Additional context

Alias sending support is also relatively important. I've personally experienced companies that wouldn't deal with me until I sent from the specific address used on my account, and didn't care for Google's internal rules for which addresses "are the same".

andrinoff:
## What?

Adds a check for `some+some@domain.com` for Gmail (sub-addressing)

## Why?

Closes #1200

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -110,9 +110,10 @@ func hasSeenFlag(flags []imap.Flag) bool {
 	return slices.Contains(flags, imap.FlagSeen)
 }
 
-// stripPlusTag removes a Gmail-style "+tag" subaddress from the local part.
-// "user+anything@gmail.com" -> "user@gmail.com".
-func stripPlusTag(addr string) string {
+// normalizeGmailAddress canonicalizes a Gmail address by stripping the "+tag"
+// subaddress and removing dots from the local part. Gmail treats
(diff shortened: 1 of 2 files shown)


andrinoff:
@LeaWhoCodes review please and thank you 🙏🏻

LeaWhoCodes (approved):
lgtm

floatpanebot (approved):

PR #1210, opened 2026-05-01
feat(oauth): drop Python helper

YeeP79:
## What?

Replaces the embedded ~500-line Python OAuth2 helper script (`config/oauth_script.py`) with a native Go implementation built on `golang.org/x/oauth2`. Token and client-credential storage moves into the OS keyring alongside the IMAP passwords already stored there.

File-by-file:

- **`config/oauth.go`** — replaced. Authorization code + PKCE (S256) flow, localhost callback on a random free port (no more hardcoded `:8189`), keyring-backed storage, transparent on-disk → keyring migration. New `OAuth2Options{StatusWriter io.Writer}` so the TUI can pass `io.Discard`. New `RevokeOAuth2Token(email)` helper.
- **`config/oauth_script.py`** — deleted, along with `OAuthScriptPath()` and the `python3` shell-out in `main.go`.
- **`main.go`** — `runOAuthCLI` rewritten to call the Go API directly. Fixed a flag-parsing bug where flags after the email positional were silently ignored (stdlib `flag` stops at the first positional; the email is now sniffed from args before `flag.Parse`). TUI launch path passes `OAuth2Options{StatusWriter: io.Discard}` to suppress stderr writes that corrupted the rendered TUI.
- **`go.mod`** — `golang.org/x/oauth2` promoted from indirect to direct.
- **Docs** — setup guides for Gmail and Outlook restructured to explain shared vs per-account Client modes with correct Google / Microsoft Entra terminology, hardcoded ports removed, `python3 not found` rows dropped from troubleshooting tables, Workspace admin-block and provider-mapping recovery rows added. CLI reference, usage doc, and `config/README.md` updated to describe the keyring schema instead of pointing at on-disk JSON files. The `matcha gmail` subcommand alias is preserved.
- **Tests** — new `config/oauth_test.go`. Coverage listed in the test plan below.

Keyring schema introduced:

- `<email>:oauth-token` — refresh + access token JSON
- `<email>:oauth-client` — per-account OAuth Client override
- `<email>:oauth-provider` — `gmail` or `outlook`, persisted so Workspace addresses on custom domains survive token refresh
- `<provider>:oauth-client` — provider-level default Client, populated on first authorization

Lookup is per-email override first, then provider default.

## Why?

Three concrete problems with the Python helper:

The single-binary distribution story breaks on OAuth. `matcha` ships via aur-bin, Homebrew, Snap, and Flatpak — all of which resolve the Go binary fine, but the moment a user enables OAuth they hit `python3 not found` or pick up an incompatible interpreter. This is the most common row in the OAuth troubleshooting table.

Credential storage is split across two backends. IMAP passwords live in the OS keyring (Secret Service / Keychain / Credential Manager). OAuth tokens lived in `~/.config/matcha/oauth_tokens/` and OAuth client credentials in `~/.config/matcha/oauth/`. Neither was covered by `secure mode`'s at-rest encryption. Consolidating on the keyring fixes both.

The old schema assumed one OAuth Client per provider. That's fine for personal accounts but breaks for Workspace, where the OAuth Client typically has to live in a Cloud Project owned by the org — meaning a personal Gmail and a Workspace Gmail can't share a Client. The new keyring layout supports per-account overrides while keeping the single-Client default for the common case.

## Migration

Existing users don't re-authenticate. On first call after upgrade:

- Tokens at `~/.config/matcha/oauth_tokens/<email>.json` are read, decoded (the old Python `expires_at` unix-timestamp shape is recognized), pushed into the keyring under `<email>:oauth-token`, and the on-disk file is deleted.
- Client credentials at `~/.config/matcha/oauth/<provider>_client.json` are read on first lookup; they're saved into the keyring properly the next time the user runs `oauth auth`.
- Workspace custom-domain accounts authorized before this PR don't have an `<email>:oauth-provider` keyring entry. Domain detection covers `@gmail.com` / `@outlook.com` etc., but custom-domain Workspace users will need to re-run `matcha oauth auth <email> --provider gmail` once. Documented in the Gmail troubleshooting table.

Implementation note for reviewers: the token decoder is a single combined struct rather than try-then-fallback because `oauth2.Token`'s `access_token` / `refresh_token` JSON tags collide with the Python shape — a native decode appears to succeed with empty `Expiry` and never falls through to the legacy parser.

## Out of scope (deliberate)

One thing I really wanted to do but decided to make a follow-up PR:

The TUI currently shows no in-flight feedback during OAuth (just the eventual `OAuth2CompleteMsg`). The Python version wrote progress to stderr, which corrupted the rendered TUI. This PR silences those writes; surfacing the auth URL and "waiting for callback" status through a proper Bubbletea modal is its own change with its own UX trade-offs.

`config/oauth_test.go` covers: RFC 7636 Appendix B PKCE S256 vector; email-domain provider detection (gmail, googlemail, outlook, hotmail, live, msn, malformed); provider resolution (auto-detect, explicit, unknown, case-insensitive); token JSON decoding for both native and legacy shapes; token round-trip through the keyring (`keyring.MockInit()`, matching the existing `config_test.go` pattern); migration from on-disk legacy tokens (verifies file is deleted after pull-into-keyring); client credentials round-trip and migration; per-email override semantics and fallback to provider default; email→provider keyring mapping with domain-detection fallback; `deleteToken` cleaning up per-email creds while leaving the provider default in place; `RevokeOAuth2Token` no-op when no token is stored.

---

*AI-assisted: implementation, tests, and initial doc drafts written collaboratively with Claude. All code reviewed line-by-line and tested end-to-end against real Google Cloud OAuth Clients (personal Gmail + Workspace).*


floatpanebot (dismissed):
Hi @YeeP79! Please fix the following issues with your PR:

- **Title**: Is too long (72 characters). The PR title must be strictly under 40 characters.

andrinoff:
@YeeP79 i'll have a look at this at the end of the day, a big PR. Thank you for your contribution!

YeeP79:
@andrinoff No problem. Awesome project, when I saw it on Reddit I knew I wanted in. I'm trying to avoid creating stuff you are asking for, but this is a legit upgrade based on the multi account issue I found. IMO of course.

andrinoff (changes requested):
The script is written and embedded with python, so that it is easily extractable and usable outside of the binary. e.g. you want to set up your provider with matcha. For that sole reason, it will remain in python and as a separate, and embedded file.

YeeP79:
Ok. I'll write something up in Python to solve it.
Thanks.

PR #1213, opened 2026-05-01
feat: markdown for html bodies

YeeP79:
## What?

Threads MIME type detection from the fetch layer through to rendering. FetchEmailBody and the IMAP/JMAP/POP3 providers now return BodyMIMEType. The Backend interface, daemon RPC protocol, and local cache schema all carry the MIME type through. view.ProcessBody takes the MIME type and skips the markdownToHTML pre-pass when it's text/html. The TUI message fetcher and view call sites are updated for the new field.

## Why?
Right now every email body gets run through md4c before we render it as HTML. The assumption was that md4c would pass raw HTML through untouched, but that falls apart on real HTML email — Datadog digests, marketing stuff, anything with heavy attributes or indentation trips md4c's html_block rules. When that happens, md4c either escapes the tags or treats them as code blocks, and the TUI ends up showing literal `<table>`, `<tr>`, `<td>` instead of the rendered content.
Detecting text/html and skipping the markdown pass fixes that, while text/plain senders like GitHub notifications and mailing lists keep their markdown formatting.

### Trade-offs & compatibility
If a text/html sender writes `**bold**` inside their HTML, it won't render as bold anymore. That's spec-correct behavior, and it's worth losing to get tables back. On the compat side, MIME type defaults to an empty string in the wire format and cache, so old cache entries fall back to the legacy markdown path. No migration needed.

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -75,18 +75,19 @@ type Attachment struct {
 }
 
 type Email struct {
-	UID         uint32
-	From        string
-	To          []string
-	ReplyTo     []string
-	Subject     string
-	Body        string
-	Date        time.Time
-	IsRead      bool
-	MessageID   string
-	References  []string
-	Attachments []Attachment
-	AccountID   string // ID of the account this email belongs to
+	UID          uint32
+	From         string
+	To           []string
+	ReplyTo      []string
+	Subject      string
+	Body         string
+	BodyMIMEType string // "text/html" or "text/plain"; empty when unknown (legacy cache rows). Lets the renderer skip markdown→HTML for already-HTML bodies.
+	Date         time.Time
+	IsRead       bool
+	MessageID    string
+	References   []string
+	Attachments  []Attachment
+	AccountID    string // ID of the account this email belongs to
 }
 
 // Folder represents an IMAP mailbox/folder.
@@ -571,15 +572,19 @@ func FetchMailboxEmails(account *config.Account, mailbox string, limit, offset u
 	return allEmails, nil
 }
 
-func FetchEmailBodyFromMailbox(account *config.Account, mailbox string, uid uint32) (string, []Attachment, error) {
+// FetchEmailBodyFromMailbox returns the chosen body, its MIME type
+// ("text/html" or "text/plain"; empty if it could not be resolved), the
+// parsed attachments, and any error. The MIME type lets the renderer
+// skip the markdown→HTML pre-pass for already-HTML bodies.
+func FetchEmailBodyFromMailbox(account *config.Account, mailbox string, uid uint32) (string, string, []Attachment, error) {
 	c, err := connect(account)
 	if err != nil {
-		return "", nil, err
+		return "", "", nil, err
 	}
 	defer c.Close()
 
 	if _, err := c.Select(mailbox, nil).Wait(); err != nil {
-		return "", nil, err
+		return "", "", nil, err
 	}
 
 	uidSet := imap.UIDSetNum(imap.UID(uid))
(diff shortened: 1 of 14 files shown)


floatpanebot (dismissed):
Hi @YeeP79! Please fix the following issues with your PR:

floatpanebot (dismissed):
Hi @YeeP79! Please fix the following issues with your PR:

andrinoff:
Hello, I cannot replicate what it fixes, can you describe in detail and also provide an email (HTML body) that is formatted better with your change?

YeeP79:
I'll get some screenshots and create a test if possible that should fail on the old setup.

andrinoff:
hey, you haven't sent an HTML body that is rendered better (you can screenshot side-by-side)

YeeP79:
Sorry for the hold up, my daughter is graduating college this weekend. It has been crazy for us. Monday before work I will knock out any issues you are seeing in either PR.

andrinoff:
> Sorry for the hold up, my daughter is graduating college this weekend. It has been crazy for us. Monday before work I will knock out any issues you are seeing in either PR.

No worries, man, take your time, congrats on the graduation!

YeeP79:
Side note on the screenshots and raw HTML: I was seeing the issues mostly in emails from my work so I need to be kind of selective on what I show. I'll make sure the unit test I create makes it crystal clear if you run it with and without the PR.

andrinoff:
> Side note on the screenshots and raw HTML: I was seeing the issues mostly in emails from my work so I need to be kind of selective on what I show. I'll make sure the unit test I create makes it crystal clear if you run it with and without the PR.

i just want to see the change, because in my inbox in the last 50 emails, i do not have a single one that is loaded differently with your fix, you can redact the information from the screenshots

andrinoff:
*updated the branch with rebase*

YeeP79:
@andrinoff 

You'll see two tests in `view/html_test.go` that share the same input (`datadogShapeHTML` — a Datadog Daily Digest-shaped table with 4-space indented attribute-heavy HTML):

- **`TestProcessBody_LegacyPathManglesIndentedHTML`** — calls `ProcessBody` with an empty MIME type (the path master takes for every email today) and asserts the rendered output literally contains `<table cellpadding=...>`. This pins the bug: md4c's html_block rule rejects indented HTML with attributes, so the markdown pre-pass leaks raw tags through to the viewer.
- **`TestProcessBody_HTMLMIMETypeSkipsMarkdownPrepass`** — same input, but with `BodyMIMETypeHTML`. Asserts the literal `<table` does not appear in the output. This pins the fix.

Both pass on this branch. If you cherry-pick the legacy assertion onto master and remove the `mimeType` parameter, the fix-side test cannot exist on master because the parameter doesn't exist there — the contract change is the fix. Run the same command from master with just the legacy test (`s/ProcessBody(..., "", ...)/ProcessBody(...)/`) and it passes there too, confirming master takes the broken path unconditionally.

fetcher/fetcher.go
```diff
@@ -950,10 +965,12 @@ func FetchEmailBodyFromMailbox(account *config.Account, mailbox string, uid uint
 									if strings.HasPrefix(ct, "text/html") {
 										body, _ := io.ReadAll(p.Body)
 										extractedBody = string(body)
+										extractedBodyMIMEType = "text/html"
 										htmlPartID = "decrypted"
 									} else if strings.HasPrefix(ct, "text/plain") && extractedBody == "" {
 										body, _ := io.ReadAll(p.Body)
 										extractedBody = string(body)
+										extractedBodyMIMEType = "text/plain"
 										htmlPartID = "decrypted"
```
andrinoff: why is this `htmlPartID`? this will result in all emails being treated as `text/html`

andrinoff (changes requested):

andrinoff (approved):
lgtm

LeaWhoCodes (approved):
looks good

PR #1214, opened 2026-05-02
fix(fetcher): decode bad charset

### LINKED ISSUE #1085: BUG: decodePart drops charset transform when Content-Type parse fails
### Describe the bug

`fetcher/fetcher.go:135-143`:

```go
mediaType, params, err := mime.ParseMediaType(header.Get("Content-Type"))
if err != nil {
    body, readErr := io.ReadAll(reader)
    if readErr != nil { ... }
    return string(body), nil
}
```

When Content-Type is malformed but a charset parameter would still have been parseable (or is implied), the fallback returns the raw bytes as a string with no charset transform. Non-UTF-8 emails (e.g. ISO-8859-1) render as mojibake.

### Expected behavior

If parse fails, attempt a best-effort charset detection (or assume UTF-8 explicitly with `transform.NewReader`) instead of returning raw bytes.

guyua9:
## What?
- Fix `decodePart` so malformed `Content-Type` headers still go through the charset decoder instead of returning raw bytes.
- Add a best-effort charset fallback for invalid media parameters.
- Add regression tests for malformed `Content-Type` with ISO-8859-1 bytes and malformed `Content-Type` without a charset.

Fixes #1085.

## Why?
Malformed `Content-Type` headers can still include a usable charset. Returning raw bytes in that path makes non-UTF-8 email bodies render as mojibake. Decoding with the best-effort charset keeps those messages readable while preserving UTF-8 fallback behavior.

## Tests
- `/Users/ming/.cache/go-toolchains/go1.26.2/bin/go test ./fetcher`
- `/Users/ming/.cache/go-toolchains/go1.26.2/bin/go vet ./fetcher`
- `/Users/ming/.cache/go-toolchains/go1.26.2/bin/go test ./fetcher ./config ./backend ./backend/jmap ./cli ./daemon ./daemonclient ./daemonrpc ./internal/httpclient ./plugin`

Note: `go test ./...` was attempted locally, but this environment could not complete unrelated packages because of a `proxy.golang.org` EOF while downloading `github.com/arran4/golang-ical` and a local macOS CommandLineTools cgo architecture error.

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -168,36 +168,52 @@ func deliveryHeadersMatch(data []byte, fetchEmail string, account *config.Accoun
 }
 
 func decodePart(reader io.Reader, header mail.PartHeader) (string, error) {
-	mediaType, params, err := mime.ParseMediaType(header.Get("Content-Type"))
-	if err != nil {
-		body, readErr := io.ReadAll(reader)
-		if readErr != nil {
-			return string(body), fmt.Errorf("fallback read after Content-Type parse error (%v): %w", err, readErr)
-		}
-		return string(body), nil
-	}
+	contentType := header.Get("Content-Type")
+	mediaType, params, parseErr := mime.ParseMediaType(contentType)
 
 	charset := "utf-8"
-	if params["charset"] != "" {
+	if parseErr != nil {
+		charset = bestEffortCharset(contentType)
+	} else if params["charset"] != "" {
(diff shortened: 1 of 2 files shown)


floatpanebot (dismissed):
Hi @guyua9! Please fix the following issues with your PR:

- **Title**: Is too long (54 characters). The PR title must be strictly under 40 characters.

andrinoff (approved):
lgtm

floatpanebot (approved):

PR #1219, opened 2026-05-03
fix(tui): aggregate folder fetch errors

### LINKED ISSUE #1126: BUG: TUI fetchFolderEmailsCmd silently drops per-account errors
### Describe the bug

`main.go:2549-2555`: same pattern as the folder fetch — errors per account are returned via `return` from the goroutine without being collected. A broken account silently shows no mail, and the user has no signal beyond an empty list.

### Expected behavior

Aggregate per-account errors into the resulting `tui.FoldersFetchedMsg` (or a sibling event) so the UI can flag the failing account.

mvanhorn:
## What?

- **`tui/messages.go`** -- `FolderEmailsFetchedMsg` gains an `AccountErrors map[string]error` field. Existing fields untouched, so this is additive for all current consumers.
- **`main.go`** -- `fetchFolderEmailsCmd` now collects per-account errors into an `accountErrors` map under the existing mutex (replacing the silent-skip return) and populates `AccountErrors` on the returned message.

## Why?

`fetchFolderEmailsCmd` (main.go) silently dropped `fetcher.FetchFolderEmails` errors per account by `return`-ing from the goroutine. A broken account showed no mail, and the UI had no signal beyond an empty list.

Closes #1126

### DIFF
--- a/main.go
+++ b/main.go
@@ -515,6 +515,11 @@ func (m *mainModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		if m.folderInbox == nil {
 			return m, nil
 		}
+		// Surface per-account fetch failures so a broken account is visible
+		// in the logs rather than silently producing an empty inbox.
+		for accID, err := range msg.AccountErrors {
+			log.Printf("folder %s: account %s fetch failed: %v", msg.FolderName, accID, err)
+		}
 		// Call plugin hooks for received emails
 		if m.plugins != nil {
 			for _, email := range msg.Emails {
@@ -2165,7 +2170,7 @@ func fetchFoldersCmd(cfg *config.Config) tea.Cmd {
 func fetchFolderEmailsCmd(cfg *config.Config, folderName string) tea.Cmd {
 	return func() tea.Msg {
 		emailsByAccount := make(map[string][]fetcher.Email)
-		accountErrors := make(map[string]error)
+		var accountErrors map[string]error
 		var mu sync.Mutex
 		var wg sync.WaitGroup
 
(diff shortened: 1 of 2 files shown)


floatpanebot (dismissed):

main.go
```diff
@@ -2696,6 +2165,7 @@ func fetchFoldersCmd(cfg *config.Config) tea.Cmd {
 func fetchFolderEmailsCmd(cfg *config.Config, folderName string) tea.Cmd {
 	return func() tea.Msg {
 		emailsByAccount := make(map[string][]fetcher.Email)
+		accountErrors := make(map[string]error)
```
andrinoff: `accountErrors` is allocated, even if there are no errors

main.go
```diff
@@ -2731,8 +2203,9 @@ func fetchFolderEmailsCmd(cfg *config.Config, folderName string) tea.Cmd {
 		}
 
 		return tui.FolderEmailsFetchedMsg{
-			Emails:     allEmails,
-			FolderName: folderName,
+			Emails:        allEmails,
+			FolderName:    folderName,
+			AccountErrors: accountErrors,
```
andrinoff, line 2216: @mvanhorn the `AccountErrors` is never used

andrinoff (changes requested):
the `AccountErrors` is never used

mvanhorn:
Pushed 817f976 addressing both inline notes:

- `accountErrors` is now declared as a nil `map[string]error` and lazy-allocated under the existing mutex on the first failure, so the success path no longer pays for an unused map (line 2168 comment).
- `AccountErrors` is now consumed: the `tui.FolderEmailsFetchedMsg` handler in `main.go` ranges over it and `log.Printf`s each per-account failure, so a broken account becomes visible in the logs instead of producing a silent empty inbox. The field carries its weight now (line 2208 comment).

Build + `go vet` + `go test -short ./...` all green locally.

andrinoff:
Bad issue, silently handling was intended, sorry.

PR #1227, opened 2026-05-03
feat(config): add LRU eviction

### LINKED ISSUE #1171: BUG: Body cache has no max size or LRU eviction
### Describe the bug

`config.SaveEmailBody` writes every fetched body into `~/.cache/matcha/email_bodies/` and never evicts. After a year of heavy use the cache hits hundreds of MB / GB; on small SSDs (laptops, Raspberry Pi) this is a real quota issue.

### Expected behavior

Configurable cap (default 500 MB?) with LRU eviction. The existing `PruneEmailBodyCache` only prunes by UID set, not by size.

### Why it's hard

LRU eviction needs an access log persisted alongside each entry, and the eviction has to be safe against concurrent reads. Doing it right means a tiny SQLite-backed metadata sidecar.

mavonx:
## What?

Implemented a size-based LRU (Least Recently Used) eviction policy for the email body cache.

## Why?

The cache grows without bound, reaching hundreds of `MB` or `GB` after heavy use. This causes disk quota issues on devices with small storage.

Closes #1171

### DIFF
--- a/config/cache.go
+++ b/config/cache.go
@@ -524,13 +524,11 @@ func evict(cache *EmailBodyCache, newSize int, threshold int) {
 }
 
 // SaveEmailBody saves or updates a cached email body for a folder.
-func SaveEmailBody(folderName string, body CachedEmailBody) error {
+func SaveEmailBody(folderName string, body CachedEmailBody, threshold int) error {
 	cache, err := LoadEmailBodyCache(folderName)
 	if err != nil {
 		cache = &EmailBodyCache{FolderName: folderName}
 	}
-
-	const threshold = 500 * 1024 * 1024 // 500MB
 
 	body.CachedAt = time.Now()
 	body.LastAccessedAt = time.Now()
--- a/config/cache.go
+++ b/config/cache.go
@@ -497,12 +497,27 @@ func GetCachedEmailBody(folderName string, uid uint32, accountID string) *Cached
 	if err != nil {
 		return nil
 	}
-	for _, b := range cache.Bodies {
+	for i, b := range cache.Bodies {
 		if b.UID == uid && b.AccountID == accountID {
-			return &b
+			cache.Bodies[i].LastAccessedAt = time.Now()
+			_ = saveEmailBodyCache(cache)
+			return &cache.Bodies[i]
 		}
 	}
 	return nil
+}
+
+func calculateEmailBodySize(body *CachedEmailBody) int {
+	size := len(body.Body)
+	for _, att := range body.Attachments {
+		size += len(att.Filename)
+		size += len(att.PartID)
+		size += len(att.Encoding)
+		size += len(att.MIMEType)
+		size += len(att.ContentID)
+		size += len(att.CalendarData)
+	}
+	return size
 }
 
 func calculateTotalCacheSize(cache *EmailBodyCache) int {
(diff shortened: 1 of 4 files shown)


floatpanebot (dismissed):

- **Title**: Is too long (50 characters). The PR title must be strictly under 40 characters.

floatpanebot (dismissed):

- **Title**: Is too long (50 characters). The PR title must be strictly under 40 characters.

andrinoff:
@Haroka-74 threshold should be configurable in the config. Let the default remain 500mb. but add it in config. Also, if you can, edit the public docs to say how to configure it and what the key in `config.json` does.

mavonx:
@andrinoff 

Sounds good, I’ll sort it out tonight.
Does the current implementation look good, or anything else I should tweak?

andrinoff:
> @andrinoff
> 
> Sounds good, I’ll sort it out tonight. Does the current implementation look good, or anything else I should tweak?

i'll check the PR later, quite busy today.

config/cache.go
```diff
@@ -503,14 +505,34 @@ func GetCachedEmailBody(folderName string, uid uint32, accountID string) *Cached
...
-func SaveEmailBody(folderName string, body CachedEmailBody) error {
+func SaveEmailBody(folderName string, body CachedEmailBody, threshold int) error {
 	cache, err := LoadEmailBodyCache(folderName)
 	if err != nil {
 		cache = &EmailBodyCache{FolderName: folderName}
 	}
 
 	body.CachedAt = time.Now()
+	body.LastAccessedAt = time.Now()
+	body.SizeBytes = len(body.Body)
```
andrinoff: this does not count `CachedAttachments`, while it should. Will lead to a much larger cache, than set in config.

config/cache.go
```diff
@@ -522,6 +544,11 @@ func SaveEmailBody(folderName string, body CachedEmailBody) error {
 		}
 	}
 	if !found {
+
+		if calculateTotalCacheSize(cache)+body.SizeBytes > threshold {
+			evict(cache, body.SizeBytes, threshold)
+		}
+
 		cache.Bodies = append(cache.Bodies, body)
```
andrinoff: If one new email is more than the `threshold`, evict will drain the cache and still append.

config/cache.go
```diff
@@ -422,11 +422,13 @@ type CachedAttachment struct {
...
-	AccountID   string             `json:"account_id"`
-	Body        string             `json:"body"`
-	Attachments []CachedAttachment `json:"attachments,omitempty"`
-	CachedAt    time.Time          `json:"cached_at"`
+	UID            uint32             `json:"uid"`
+	AccountID      string             `json:"account_id"`
+	Body           string             `json:"body"`
+	Attachments    []CachedAttachment `json:"attachments,omitempty"`
+	CachedAt       time.Time          `json:"cached_at"`
+	LastAccessedAt time.Time          `json:"last_accessed_at"`
```
andrinoff, line 430: As far as I see, this is never updated. Should be updated on `GetCachedEmailBody`.

mavonx:
@andrinoff Done.

config/cache.go
```diff
@@ -522,6 +544,11 @@ func SaveEmailBody(folderName string, body CachedEmailBody) error {
 		}
 	}
 	if !found {
```
andrinoff, line 561: The least of my concerns, but also worth noting, that, if the email is changed (`found=true` still) the cache is not going to be evicted

andrinoff (changes requested):

andrinoff (approved):

floatpanebot (approved):

PR #1233, opened 2026-05-05
fix(sender): fail closed on rand error

### LINKED ISSUE #1127: BUG: SMIME outer boundary fallback is predictable on rand failure
### Describe the bug

`sender/sender.go:281-288`:

```go
var rb [12]byte
if _, rerr := rand.Read(rb[:]); rerr == nil {
    outerBoundary = "signed-" + fmt.Sprintf("%x", rb[:])
} else {
    outerBoundary = "signed-" + fmt.Sprintf("%d", time.Now().UnixNano())
}
```

If `crypto/rand` fails (rare but possible inside containers/seccomp), we fall back to UnixNano — predictable, and the boundary token is essentially attacker-knowable. Worst case: a crafted body part can collide with the fallback boundary and chop the message.

### Expected behavior

Treat `crypto/rand` failure as fatal: return an error rather than degrading to a predictable token.

SAY-5:
Closes #1127.

## What?

Treats `crypto/rand` failure as fatal in `sender.go` instead of falling back to a `time.UnixNano()`-based outer boundary. Replaces both the signed-only and signed+encrypted call sites with a `smimeOuterBoundary()` helper that propagates the error, plus regression tests using an injectable `randReader`.

## Why?

The previous fallback produced a predictable boundary that could be collided by an attacker-controlled body part, breaking S/MIME envelope integrity. Failing closed on rand errors removes the predictable path entirely; tests pin the new behavior so the fallback can't silently return.

### DIFF
--- a/sender/sender.go
+++ b/sender/sender.go
@@ -9,6 +9,7 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"io"
 	"mime"
 	"mime/multipart"
 	"mime/quotedprintable"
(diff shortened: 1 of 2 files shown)


floatpanebot (dismissed):
Hi @SAY-5! Please fix the following issues with your PR:

- **Title**: Is too long (60 characters). The PR title must be strictly under 40 characters.

andrinoff (approved):
lgtm

PR #1243, opened 2026-05-06
fix(tui): stop attachment nav scroll

FromSi:
## What?

Prevents email body scrolling when navigating attachments in the email view.

## Why?

When attachment focus is active, up/down should only move the attachment selection. Previously, those key events also reached the email body viewport, causing the message content to scroll at the same time.


andrinoff (approved):
lgtm

PR #1252, opened 2026-05-07
feat: show attachment file sizes

### LINKED ISSUE #550: FEAT: Show attachment sizes in Composer
When attaching files in the composer, only the filename is shown. It's useful to know the size of the attachments before sending.

Display the size of each attachment (e.g. "image.jpg (1.2 MB)") in the attachment list in the composer.

Checking the file size in a file manager before attaching.

### Additional context
Implementation in `tui/composer.go`.

FromSi:
Closes #550

## What?

Display attachment file sizes in the composer attachment list, e.g. `image.jpg (1.2 MB)`.

Also wired file size formatting through the existing i18n number formatter:

- added `GetNumberFormatter()` to the i18n manager
- added a short TUI helper `tfs(...)`
- reused it in both composer and file picker
- removed the old file picker-local file size formatter
- added tests for attachment display

## Why?

Composer previously showed only attachment filenames, so users had no quick way to check file size before sending.

The formatter was touched because there was already an unused locale-aware `i18n.NumberFormatter.FormatFileSize`. Instead of keeping a composer-specific helper or reusing a file picker-private function, this change makes file size formatting a shared i18n-backed path. That keeps formatting consistent across TUI components and supports locale-specific number formatting, such as `1.2 MB` vs `1,2 MB`.

### DIFF
--- a/tui/composer.go
+++ b/tui/composer.go
@@ -2,6 +2,7 @@ package tui
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 
@@ -228,6 +229,15 @@ func (m *Composer) getSelectedAccount() *config.Account {
 		return &m.accounts[m.selectedAccountIdx]
 	}
 	return nil
+}
+
+func formatAttachmentName(path string) string {
+	name := filepath.Base(path)
+	info, err := os.Stat(path)
+	if err != nil || info.IsDir() {
+		return name
+	}
+	return fmt.Sprintf("%s (%s)", name, tfs(info.Size()))
 }
 
 func (m *Composer) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
@@ -607,7 +617,7 @@ func (m *Composer) View() tea.View {
 	} else {
 		var names []string
 		for _, p := range m.attachmentPaths {
-			names = append(names, filepath.Base(p))
+			names = append(names, formatAttachmentName(p))
 		}
 		attachmentText := strings.Join(names, ", ")
 		if m.focusIndex == focusAttachment {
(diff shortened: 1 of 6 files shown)


FromSi:
Closes #550

andrinoff:
> Closes #550

add it to description, please.

tui/composer.go
```diff
@@ -607,7 +617,7 @@ func (m *Composer) View() tea.View {
 	} else {
 		var names []string
 		for _, p := range m.attachmentPaths {
-			names = append(names, filepath.Base(p))
+			names = append(names, formatAttachmentName(p))
```
andrinoff: this will possibly cause performance issues, as View() gets called on every press/window resize. The file size should be cached at the point the attachment is added, not re-queried at render time.

andrinoff (changes requested):

tui/composer.go
FromSi: Good point, I agree. Would it make sense to move this into `Update()`? Or are there better places?

andrinoff: @FromSi a simple map[string]string keyed by path (populated in the key handler when `PickedFileMsg` is processed) should fix this.

FromSi: I couldn’t find a `PickedFileMsg` here, so I assumed you meant the existing `FileSelectedMsg`. I populated the cache there, since that’s the message currently emitted by the file picker when attachments are selected.

andrinoff (approved):
lgtm

andrinoff:
i'll check how it looks in the evening.

tui/composer.go
andrinoff: yeah, thats what it was, sorry

floatpanebot (approved):

PR #1266, opened 2026-05-09
feat: plugin settings SDK

andrinoff:
## What?

Adds user-configurable plugin settings. Plugins declare typed options via a new `matcha.settings(spec)` Lua API supporting `boolean`, `number`, and `string` types with defaults, labels, and descriptions. The call returns a read-only proxy table whose fields reflect live values, so hook callbacks read current settings through closure capture. A second helper `matcha.get_setting(key [, plugin])` provides explicit lookup. The plugin Manager tracks the loading plugin so schemas attribute correctly, and gains methods for reading, writing, loading, and exporting setting values with type coercion. Values persist in `config.json` under a new `plugin_settings` field (both regular and secure-mode disk formats).

## Why?

Plugins shipped hardcoded constants (keyword lists, thresholds, toggles). Users had to edit Lua source to tweak them — fragile and lost on plugin reinstall. Plugins now declare schemas; users edit values in TUI; values persist across upgrades. Bool toggle avoids parser errors from typing `true`/`false`/`1`/`0` inconsistently.

### DIFF
--- a/plugin/settings.go
+++ b/plugin/settings.go
@@ -1,0 +1,315 @@
+package plugin
+
+import (
+	"sort"
+
+	lua "github.com/yuin/gopher-lua"
+)
+
+// SettingType identifies the kind of value a plugin setting holds.
+type SettingType string
+
+const (
+	SettingBool   SettingType = "boolean"
+	SettingNumber SettingType = "number"
+	SettingString SettingType = "string"
+)
+
+// SettingDef describes a single configurable plugin setting.
+type SettingDef struct {
+	Key         string
+	Type        SettingType
+	Default     interface{}
+	Label       string
+	Description string
+}
+
+// PluginSettings holds the schema for one plugin's settings, ordered by
+// declaration order so the TUI lists them predictably.
+type PluginSettings struct {
+	Plugin string
+	Defs   []SettingDef
+}
+
+// declareSettings registers the schema for the currently-loading plugin and
+// returns a Lua table whose __index reads the live current value for a key.
+// Plugins typically capture the returned table in a local and read fields
+// from it at any later time, including inside hook callbacks:
+//
+//	local cfg = matcha.settings({
+//	    threshold = {type = "number", default = 5},
+//	    enabled   = {type = "boolean", default = true},
+//	})
+//	matcha.on("email_received", function(email)
+//	    if cfg.enabled and #email.subject > cfg.threshold then ... end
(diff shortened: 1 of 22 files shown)


tui/settings_plugins.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+func (m *Settings) updatePluginList(msg tea.KeyPressMsg) (tea.Model, tea.Cmd) {
+	schemas := m.plugins.Schemas()
+	if len(schemas) == 0 {
+		return m, nil
+	}
+
+	switch msg.String() {
+	case "up", "k":
```
FromSi: Maybe we should add a keybind for `settings_plugins`? :))

andrinoff: `k` and `j` are configured from the `global` keybinds, so i'll add them, arrow keys are there by default and will remain unchanged

plugin/settings.go
```diff
@@ -0,0 +1,315 @@
...
+func (m *Manager) lookupValue(plugin, key string, def SettingDef) interface{} {
+	if vals, ok := m.pluginValues[plugin]; ok {
+		if v, ok := vals[key]; ok {
+			return v
+		}
+	}
+	return def.Default
+}
+
+func toLuaValue(L *lua.LState, v interface{}) lua.LValue {
```
FromSi, line 180: Maybe it makes sense to move `toLuaValue` into a small internal helper package, e.g. `internal/luautil`.

This conversion is not really specific to plugin settings, and we already have similar places where we manually construct Lua values. For example, `plugin/hooks.go` currently does:

```go
t.RawSetString("is_read", lua.LBool(isRead))
```

With a shared helper this could become:

```go
t.RawSetString("is_read", luautil.ToValue(isRead))
```

So `plugin/settings.go` would use:

```go
L.Push(luautil.ToValue(val))
```

and hooks could reuse the same conversion when building Lua tables from Go values. Something like:

```go
package luautil

func ToValue(v any) lua.LValue {
    switch x := v.(type) {
    case bool:
        return lua.LBool(x)
    case float64:
        return lua.LNumber(x)
    case int:
        return lua.LNumber(x)
    case string:
        return lua.LString(x)
    default:
        return lua.LNil
    }
}
```

The `L *lua.LState` argument is not needed for bool/number/string today.

steveevansdev (approved):
lgtm

PR #1267, opened 2026-05-10
fix(tui): surface fetchFolders errors

### LINKED ISSUE #1125: BUG: TUI fetchFoldersCmd silently drops per-account errors
### Describe the bug

`main.go:2511-2516`:

```go
go func(acc config.Account) {
    defer wg.Done()
    folders, err := fetcher.FetchFolders(&acc)
    if err != nil { return }
    ...
}(account)
```

Per-account errors are dropped on the floor. If an account's IMAP login is broken, the user sees the folder list silently miss those folders and never gets a notification.

### Expected behavior

Collect per-account errors and surface them as a non-fatal warning in the TUI status line.

mvanhorn:
## What?

Per-account `fetcher.FetchFolders` errors are now collected into
`FoldersFetchedMsg.Errors` and surfaced as a transient overlay in the
TUI (4s, then auto-restore via the existing `PluginNotifyMsg` pattern).
Closes #1125.

## Why?

`fetchFoldersCmd` spawns one goroutine per account, each calling
`fetcher.FetchFolders`. The previous code returned immediately on err
and the goroutine vanished without leaving a trace. If an account's
IMAP login was broken, OAuth had expired, or the server was
unreachable, the affected account silently dropped out of the merged
folder list and the user got no signal -- their folder list quietly
missed entries. The reporter described this exactly: "user sees the
folder list silently miss those folders and never gets a notification."

## How

| File | Change |
| --- | --- |
| `tui/messages.go` | New `Errors map[string]error` field on `FoldersFetchedMsg`. |
| `main.go` (producer at fetchFoldersCmd) | Capture per-account fetch errors into a sibling map under the existing mutex; pass through to the message. |
| `main.go` (consumer at FoldersFetchedMsg case) | When `Errors` is non-empty, surface a transient overlay listing affected accounts and their errors. |

The producer change is a 7-line localized capture: same goroutine, same
mutex, same return-after-error path -- the only difference is that the
err is stored before returning instead of being thrown away.

The consumer surfaces errors using the same pattern the codebase already
uses for `PluginNotifyMsg`: save `m.previousModel`, swap `m.current` to
a `tui.NewStatus(...)` overlay, fire `tea.Tick(4*time.Second, ...)` that
returns `RestoreViewMsg`. After 4 seconds the user is back to where
they were. No new TUI primitives, no new message types beyond the field
on the existing message.

Account display name lookup prefers `Email` -> `Name` -> `ID` so
partially configured accounts still show something meaningful in the
overlay. Errors are sorted alphabetically so the rendered list is
deterministic across restarts (the previous, per-iteration map order
would jitter).

The disjoint property (an account is in either `FoldersByAccount` *or*
`Errors`, never both) is preserved by the existing `return` after
recording the error.

## Verification

```
make lint     # go fmt ./... + go vet ./...; clean
make test     # all packages pass
go build ./... # clean
```

I did not add automated tests for the TUI surface itself -- the existing
test suite has no tests for `main.go`'s message handlers (verified via
`find . -name '*_test.go'`), so introducing the first one felt like
scope creep for a fix PR. Happy to add a coverage test for the
disjoint-by-construction property in `fetchFoldersCmd` if you'd like;
just point me at the testing pattern you'd want.

### DIFF
--- a/config/cache.go
+++ b/config/cache.go
@@ -495,6 +495,8 @@ func saveEmailBodyCache(cache *EmailBodyCache) error {
 }
 
 // GetCachedEmailBody returns the cached body for a specific email, or nil if not cached.
+// LastAccessedAt is updated by SaveEmailBody, not here -- a read should not
+// mutate cache state.
 func GetCachedEmailBody(folderName string, uid uint32, accountID string) *CachedEmailBody {
 	cache, err := LoadEmailBodyCache(folderName)
 	if err != nil {
@@ -502,8 +504,6 @@ func GetCachedEmailBody(folderName string, uid uint32, accountID string) *Cached
 	}
 	for i, b := range cache.Bodies {
 		if b.UID == uid && b.AccountID == accountID {
-			cache.Bodies[i].LastAccessedAt = time.Now()
-			_ = saveEmailBodyCache(cache)
 			return &cache.Bodies[i]
 		}
 	}
--- a/main.go
+++ b/main.go
@@ -524,6 +524,44 @@ func (m *mainModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			}
 			go config.SaveAccountFolders(accID, names)
 		}
+		// Per-account fetch errors (e.g. broken IMAP login, unreachable
+		// server) are non-fatal: other accounts' folders are still shown.
+		// Surface them as a transient overlay so the user knows why an
+		// account's folders are missing instead of silently dropping them.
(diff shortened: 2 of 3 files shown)


floatpanebot (dismissed):

- **Title**: Is too long (78 characters). The PR title must be strictly under 40 characters.

andrinoff (approved):
lgtm

mvanhorn:
Appreciate the merge, @andrinoff. fetchFolders surfacing errors beats silent failure.

PR #1268, opened 2026-05-10
fix: protect providers map

### LINKED ISSUE #766: BUG: Provider map race condition in main model
**Describe the bug**
Provider map at main.go:130-145 accessed without locking in getProvider() while other goroutines modify in initProviders(). Classic race condition.

**To reproduce**
1. Build with -race flag
2. Run matcha with multiple accounts
3. Race detector reports concurrent map access

**Expected behavior**
Use sync.RWMutex to protect provider map access

**Matcha version**
master

**OS**
All

**Additional context**
main.go lines 130-145
Concurrent map access without sync

mavonx:
## What?

Protected the `providers` map with a `sync.RWMutex` to fix a race condition.

## Why?

`ensureProviders()` writes to the map on the Bubble Tea event loop while `getProvider()` is called concurrently from goroutines spawned by batch and search commands, causing unsafe concurrent map access.

Closes #766

### DIFF
--- a/main.go
+++ b/main.go
@@ -178,12 +178,15 @@ func (m *mainModel) ensureProviders() {
 
 // getProvider returns the backend provider for the given account.
 func (m *mainModel) getProvider(acct *config.Account) backend.Provider {
-	m.providersMu.RLock()
-	defer m.providersMu.RUnlock()
 	if acct == nil {
 		return nil
 	}
-	return m.providers[acct.ID]
+
+	m.providersMu.RLock()
+	p := m.providers[acct.ID]
+	m.providersMu.RUnlock()
+
+	return p
 }
 
 func (m *mainModel) Init() tea.Cmd {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -155,20 +156,30 @@ func (m *mainModel) ensureProviders() {
...
+		m.providersMu.Lock()
 		m.providers[acct.ID] = p
+		m.providersMu.Unlock()
 	}
 }
 
 // getProvider returns the backend provider for the given account.
 func (m *mainModel) getProvider(acct *config.Account) backend.Provider {
+	m.providersMu.RLock()
+	defer m.providersMu.RUnlock()
```
andrinoff: @mavonx  shouldn't this be after checking the `acct == nil`?

mavonx: You are right, I will handle this right now.

andrinoff (approved):
lgtm

PR #1275, opened 2026-05-11
docs: Open Collective funding option

andrinoff:
## What?

Adds Opem Collective URL

## Why?

For sponsoring


floatpanebot (approved):

PR #1276, opened 2026-05-11
fix(fetcher): nil-safe charset lookup

### LINKED ISSUE #896: BUG: IANA encoding lookup error ignored for charset fallback
### Describe the bug

In fetcher/fetcher.go:152, the IANA encoding lookup error is ignored:
```go
encoding, _ = ianaindex.IANA.Encoding("utf-8")
```

While UTF-8 lookup is unlikely to fail, ignoring the error means `encoding` could be nil, leading to a nil pointer dereference when it's used to create a decoder.

### To reproduce

Unlikely in practice since UTF-8 is always registered, but violates Go best practices for error handling.

### Expected behavior

Check the error and handle it, even if just with a panic for impossible conditions.

SAY-5:
## What?

`decodeReaderWithCharset` and `decodeHeader`'s `CharsetReader` in `fetcher/fetcher.go` could nil-deref because `ianaindex.IANA.Encoding` is allowed to return `(nil, nil)` for charset names it recognizes but has no implementation for. The previous code chained `.NewDecoder()` on that nil encoding.

Refactor: introduce `lookupCharsetEncoding` that always returns a non-nil `encoding.Encoding` by trying the requested charset, then `ianaindex`'s "utf-8", then `encoding/unicode.UTF8` as final guarantee. Use it from `decodeReaderWithCharset`. Guard the `decodeHeader` path explicitly so the nil-after-success case surfaces as a clean error.

## Why?

Fixes #896.

Without the guard, a real message with an unusual charset header can panic the fetcher. The fix is minimal, has no behavior change for the common path (a valid charset still resolves the same way), and is covered by two new tests:

- `TestDecodeReaderWithCharsetSurvivesUnknownCharset` (issue regression)
- `TestLookupCharsetEncodingAlwaysReturnsNonNil` (table-driven contract)

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -30,7 +30,9 @@ import (
 	"github.com/emersion/go-pgpmail"
 	"github.com/floatpane/matcha/config"
 	"go.mozilla.org/pkcs7"
+	"golang.org/x/text/encoding"
 	"golang.org/x/text/encoding/ianaindex"
+	"golang.org/x/text/encoding/unicode"
 	"golang.org/x/text/transform"
 )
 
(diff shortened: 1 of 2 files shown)


floatpanebot (dismissed):
Hi @SAY-5! Please fix the following issues with your PR:

floatpanebot (dismissed):
Hi @SAY-5! Please fix the following issues with your PR:

andrinoff (approved):
lgtm

PR #1277, opened 2026-05-12
fix: mailing-list addresses in a slice

### LINKED ISSUE #1123: BUG: Mailing-list virtual contact stores joined addresses in Email field
### Describe the bug

`config/cache.go:217-225`:

```go
matches = append(matches, Contact{
    Name:     list.Name,
    Email:    strings.Join(list.Addresses, ", "),
    UseCount: 9999,
    LastUsed: time.Now(),
})
```

The `Email` field is supposed to hold a single address. Storing a comma-joined string of addresses there means downstream code that calls `normalizeContactEmail` lowercases and trims commas, corrupting the data — and any code that does an exact-match lookup against this contact will never find it.

### Expected behavior

Either model mailing lists with a separate type, or store the addresses in a dedicated `Addresses []string` field and keep `Email` empty / set to the list name.

mvanhorn:
## What?

`SearchContacts` builds mailing-list virtual contacts by jamming `strings.Join(list.Addresses, ", ")` into `Contact.Email`. That breaks the field's single-address invariant: `normalizeContactEmail` lowercases and trims commas off the joined string (corrupting the data), and any exact-match lookup against the contact will never find it.

This adds a dedicated `Addresses []string` field to `Contact`, populates it for mailing-list virtual contacts, and leaves `Email` empty so the invariant holds. The composer learns to detect mailing lists via `len(selected.Addresses) > 0` instead of "does the email field contain a comma", joins the addresses at insertion time, and renders the suggestion as `Name (addr1, addr2, ...)` so it's clear that a single suggestion expands to multiple recipients.

Files touched:

- `config/cache.go` -- new `Addresses` field on `Contact`; mailing-list construction in `SearchContacts` switches to it
- `tui/composer.go` -- detection/insertion and rendering swap from `Email` commas to `Addresses`

The cache file format is backwards-compatible: the new field is `omitempty`, and saved contacts (added via `AddContact`) never carry `Addresses` because they're built from a single email parameter. Mailing-list virtual contacts are constructed in-memory on every search and never written to disk.

## Why?

Closes #1123. The reporter pointed out the data-integrity bug; this implements option 2 of the two fixes they suggested (dedicated `Addresses` slice, keep `Email` empty), which is the smaller of the two and doesn't require a separate "virtual contact" type.

## Verification

- `go build ./config/... ./tui/...`: passes
- `go vet ./config/... ./tui/...`: passes
- `go test ./config/...`: 51/51 passing
- `go test ./tui/...`: 79/79 passing

## Disclosure

Authored with Claude as a coding assistant; I reviewed the change before pushing.

### DIFF
--- a/tui/composer.go
+++ b/tui/composer.go
@@ -275,6 +275,38 @@ func (m *Composer) removeSelectedAttachment() {
 	m.clampAttachmentCursor()
 }
 
+func suggestionDisplay(s config.Contact, suggestionWidth int) string {
+	display := s.Email
+	if len(s.Addresses) > 0 {
+		display = fmt.Sprintf("%s (%s)", s.Name, strings.Join(s.Addresses, ", "))
+		return truncateSuggestionDisplay(display, suggestionWidth)
+	} else if s.Name != "" && s.Name != s.Email {
+		display = fmt.Sprintf("%s <%s>", s.Name, s.Email)
+	}
+	return display
+}
+
+func suggestionDisplayWidth(width int) int {
+	if width > 12 {
+		return width - 6
+	}
+	return 40
+}
+
+func truncateSuggestionDisplay(s string, maxLen int) string {
+	runes := []rune(s)
+	if len(runes) <= maxLen {
+		return s
+	}
+	if maxLen <= 0 {
+		return ""
+	}
+	if maxLen <= 3 {
+		return string(runes[:maxLen])
+	}
+	return string(runes[:maxLen-3]) + "..."
+}
+
 func (m *Composer) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	var cmds []tea.Cmd
 	var cmd tea.Cmd
@@ -699,13 +731,9 @@ func (m *Composer) View() tea.View {
 	toFieldView := m.toInput.View()
 	if m.showSuggestions && len(m.suggestions) > 0 {
 		var suggestionsBuilder strings.Builder
+		suggestionWidth := suggestionDisplayWidth(m.width)
 		for i, s := range m.suggestions {
-			display := s.Email
-			if len(s.Addresses) > 0 {
-				display = fmt.Sprintf("%s (%s)", s.Name, strings.Join(s.Addresses, ", "))
-			} else if s.Name != "" && s.Name != s.Email {
-				display = fmt.Sprintf("%s <%s>", s.Name, s.Email)
-			}
+			display := suggestionDisplay(s, suggestionWidth)
 			if i == m.selectedSuggestion {
 				suggestionsBuilder.WriteString(selectedSuggestionStyle.Render("> "+display) + "\n")
 			} else {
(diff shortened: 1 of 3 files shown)


floatpanebot (dismissed):

- **Title**: Is too long (78 characters). The PR title must be strictly under 40 characters.

floatpanebot (dismissed):

- **Title**: Is too long (78 characters). The PR title must be strictly under 40 characters.

mvanhorn:
Resolved and pushed a29f108. The conflict was structural -- since this PR opened, master refactored `Contact` to track usage per-account (`Usage map[string]ContactUsage` with a `legacyContactUsageKey` migration path), and that bumped into the new `Addresses []string` field here.

Merged shape:

```go
type Contact struct {
    Name      string                  `json:"name"`
    Email     string                  `json:"email"`
    Addresses []string                `json:"addresses,omitempty"`
    Usage     map[string]ContactUsage `json:"usage_by_account"`
}
```

Mailing-list virtual contacts in `SearchContacts` now leave `Email` empty, copy addresses into `Addresses`, and set the 9999 usage score under the active account in the Usage map. The composer's `len(selected.Addresses) > 0` check stays as-is and is the canonical way to distinguish mailing-list vs regular contacts.

`go build ./...`, `go vet ./...`, and `go test ./config/... ./tui/... ./cli/...` all clean locally. Branch shows MERGEABLE.

floatpanebot (dismissed):

- **Title**: Is too long (44 characters). The PR title must be strictly under 40 characters.

tui/composer.go
```diff
@@ -701,7 +701,9 @@ func (m *Composer) View() tea.View {
 		var suggestionsBuilder strings.Builder
 		for i, s := range m.suggestions {
 			display := s.Email
-			if s.Name != "" && s.Name != s.Email {
+			if len(s.Addresses) > 0 {
+				display = fmt.Sprintf("%s (%s)", s.Name, strings.Join(s.Addresses, ", "))
```
andrinoff: `fmt.Sprintf("%s (%s)", s.Name, strings.Join(s.Addresses, ", "))` for a mailing list with many recipients could produce a very long suggestion string and overflow the TUI layout. Let's truncate

andrinoff (changes requested):
Fix the issue below, and add tests

mvanhorn:
Addressed in 14a67a6 — extracted a `suggestionDisplay()` helper that truncates the mailing-list branch to the composer's available width (`width - 6`, 40-rune fallback). Truncation is rune-safe and adds an ellipsis. Single-address suggestions are untouched.

Added `TestMailingListSuggestionTruncates` covering both the truncated mailing-list case and the unchanged single-address case. `go build ./...` and `go test ./tui/...` are clean.

andrinoff (approved):
lgtm

mvanhorn:
Thanks for the merge, @andrinoff. Glad the mailing-list slice fix landed cleanly.

mvanhorn:
@andrinoff thanks for the merge. Switching mailing-list virtual contacts to a slice instead of jamming `strings.Join(list.Addresses, ", ")` into a single Email field preserves the normalizeContact invariant cleanly.

Issue #448, opened 2026-04-04
BUG: E-Mail Account needs username field


BenjaminKobjolke:
### Describe the bug

Some email providers have a username which is not equal to the email address.

### To reproduce

config.json
if I use my username in "email": "USERNAME",
receiving emails works.

But when I try to send an email I get an error
Failed to send email: 504 5.5.2 <USERNAME>: Sender address rejected: need fully-qualified address

If i use my email "email": "email@domain.com"
then fetching emails doesnt work.

### Expected behavior

seperate username field for emaiil accounts

### OS

Windows 11

### Terminal

wezterm

PR #1279, opened 2026-05-12
fix(reply): use send-as envelope

### LINKED ISSUE #1274: BUG: Replying to a mailing list uses the list's address as the SMTP MAIL FROM (Causes 554 Error)
### Describe the bug

**Description :** 
When replying to an email received from a distribution/mailing list, Matcha incorrectly sets the SMTP envelope sender (MAIL FROM) to the address of the mailing list, rather than the configured email address of the user's account.

This causes the outgoing SMTP server (e.g., Postfix) to reject the message with a 554 error, as the user is not authenticated/authorized to send emails on behalf of the mailing list domain.

**Actual Behavior :**

Matcha attempts to send the email using the mailing list's address as the `MAIL FROM`. The SMTP server rejects the transaction.

Error from the logs:
`2026/05/11 11:49:00 Failed to send email: 554 Transaction failed. Please report this error to your system administrator (223E)`

### To reproduce

1. Receive an email sent via a mailing list or distribution list.
2. Click "Reply" to answer the email.
3. Send the email.
4. The email fails to send, and an error is logged.

### Expected behavior

When composing any reply, Matcha should strictly force the SMTP `MAIL FROM` (envelope sender) to match the user's configured account email address, regardless of the `Return-Path`, `Sender`, or `From` headers of the original incoming email.

### Screenshots

Image

### Matcha version

0.37.0

### Go version

1.25.9

### OS

Linux

### OS version

Fedora 43

### Architecture

amd64 (x86_64)

### Terminal

Kitty

### Additional context

_No response_

FromSi:
## What?

Updates the SMTP envelope sender for outgoing mail.

Matcha now uses `send_as_email` for `MAIL FROM` when it is configured. If it is not configured, it keeps the existing fallback behavior.

This applies to regular email sending and calendar replies.

## Why?

Closes #1274

Previously, `MAIL FROM` was based on `fetch_email`, with `account.email` as the fallback.

Now the order is:

```text
send_as_email -> fetch_email -> account.email
```

`fetch_email` is used to match incoming messages to an account. For outgoing mail, the configured Send-As address should be used first.

### DIFF
--- a/sender/sender.go
+++ b/sender/sender.go
@@ -711,7 +711,7 @@ func SendEmail(account *config.Account, to, cc, bcc []string, subject, plainBody
 	}
 
 	// Send Envelope
-	if err = c.Mail(account.Email); err != nil {
+	if err = c.Mail(account.GetSendAsEmail()); err != nil {
 		return nil, err
 	}
 	for _, r := range allRecipients {
@@ -916,7 +916,7 @@ func SendCalendarReply(account *config.Account, to []string, subject, plainBody 
 		}
 	}
 
-	if err = c.Mail(account.Email); err != nil {
+	if err = c.Mail(account.GetSendAsEmail()); err != nil {
 		return nil, err
 	}
 	for _, r := range to {
--- a/sender/sender.go
+++ b/sender/sender.go
@@ -711,7 +711,7 @@ func SendEmail(account *config.Account, to, cc, bcc []string, subject, plainBody
 	}
 
 	// Send Envelope
-	if err = c.Mail(account.GetFetchEmail()); err != nil {
+	if err = c.Mail(account.Email); err != nil {
 		return nil, err
 	}
 	for _, r := range allRecipients {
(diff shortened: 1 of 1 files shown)


andrinoff:
This is an impossible situation, where `FetchEmail` does not have the right to send on behalf of `EmailAddress` (or should be impossible)

`EmailAddress`, on the other hand can be not an email. This is why it is using `FetchEmail`


#448, change made in #449

andrinoff:
@FromSi please, change this PR, to send the envelope at first from `Send-as` address, if set, otherwise, fall back to `FetchEmail` (which will never be empty, so no need to fall back to `EmailAddress`

FromSi:
@andrinoff I updated the `SMTP` envelope sender to use `GetSendAsEmail()`, so `MAIL FROM` now uses the configured `send-as` address first and falls back through the existing `fetch-email` behavior.

Is this what you had in mind?

Also, should I remove the “Closes #...” text from the PR description?

andrinoff:
Yes, that's what I had in mind, thank you. 

No, don't remove it, it's still the fix for this.

andrinoff:
the issue creator doesn't seem to respond. I'll merge either way

andrinoff (approved):
lgtm now

andrinoff:
@FromSi can you change the PR description?

FromSi:
@andrinoff the title and description have been updated

andrinoff:
> @andrinoff the title and description have been updated

@FromSi , I think @LeaWhoCodes emailed you and invited to the org

PR #1280, opened 2026-05-12
feat: in-memory LRU email body cache

mavonx:
## What?

Replaces the disk-only email body cache with an in-memory LRU cache backed by write-through disk persistence.

- Single global LRU instance shared across all folders via `sync.Once` singleton.
- Fast `Get()` and `Put()` operations — `O(1)` via hashmap + doubly-linked list.
- Write-through: every `Get()` and `Put()` immediately persists to disk, so no data is lost on crash.
- Eviction removes least-recently-used entries from both memory and disk atomically.
- On the first `GetLRUInstance()` call, `LoadFromDisk()` restores LRU order using `LastAccessedAt` timestamps sorted oldest-first, so the most recently accessed email ends up at the front.

## Why?

The previous disk-only approach had two bottlenecks: `GetCachedEmailBody()` read the entire folder JSON file on every call, and `pruneEmailBodyCacheSize()` loaded all folder files on every `SaveEmailBody()` call. With multiple accounts and many folders, this becomes expensive.

Related #1171

### DIFF
--- a/config/cache.go
+++ b/config/cache.go
@@ -2,7 +2,6 @@ package config
 
 import (
 	"encoding/json"
-	"errors"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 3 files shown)


andrinoff:
@mavonx please, don't close PRs before they are reviewed. This was just a warning

mavonx:
@andrinoff

I opened a new one. Sorry about that.

PR #1281, opened 2026-05-12
feat: in-memory LRU email body cache

mavonx:
## What?

Replaces the disk-only email body cache with an in-memory LRU cache backed by write-through disk persistence.

- Single global LRU instance shared across all folders via `sync.Once` singleton.
- Fast `Get()` and `Put()` operations — `O(1)` via hashmap + doubly-linked list.
- Write-through: every `Get()` and `Put()` immediately persists to disk, so no data is lost on crash.
- Eviction removes least-recently-used entries from both memory and disk atomically.
- On the first `GetLRUInstance()` call, `LoadFromDisk()` restores LRU order using `LastAccessedAt` timestamps sorted oldest-first, so the most recently accessed email ends up at the front.

## Why?

The previous disk-only approach had two bottlenecks: `GetCachedEmailBody()` read the entire folder JSON file on every call, and `pruneEmailBodyCacheSize()` loaded all folder files on every `SaveEmailBody()` call. With multiple accounts and many folders, this becomes expensive.

Related #1171

### DIFF
--- a/config/lru.go
+++ b/config/lru.go
@@ -23,6 +23,7 @@ type LRU struct {
 	currentSize int
 	cache       map[string]*list.Element
 	ll          *list.List
+	mu          sync.Mutex
 }
 
 var lru *LRU
@@ -42,6 +43,9 @@ func GetLRUInstance(threshold int) *LRU {
 			}
 
 		})
+
+	lru.mu.Lock()
+	defer lru.mu.Unlock()
 
 	if lru.threshold != threshold {
 		lru.threshold = threshold
@@ -211,6 +215,9 @@ func saveEmailBodyToDisk(folder string, body *CachedEmailBody) error {
 }
 
 func (lru *LRU) Get(folder string, uid uint32, accountID string) *Node {
+	lru.mu.Lock()
+	defer lru.mu.Unlock()
+
 	key := lru.makeKey(folder, uid, accountID)
 
 	e, ok := lru.cache[key]
@@ -240,6 +247,9 @@ func (lru *LRU) removeKey(key string) {
 }
 
 func (lru *LRU) Put(folder string, uid uint32, accountID string, body *CachedEmailBody) {
+	lru.mu.Lock()
+	defer lru.mu.Unlock()
+
 	key := lru.makeKey(folder, uid, accountID)
 
 	if body.SizeBytes > lru.threshold {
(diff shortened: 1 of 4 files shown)


andrinoff:
@mavonx tests fail

mavonx:
@andrinoff 

I will double-check this again tonight.

config/lru.go
```diff
@@ -0,0 +1,267 @@
...
+var once sync.Once
+
+func GetLRUInstance(threshold int) *LRU {
+	once.Do(
+		func() {
+			lru = &LRU{
+				threshold: threshold,
+				cache:     make(map[string]*list.Element),
+				ll:        list.New(),
+			}
```
andrinoff, line 39: this will set the threshold only once, this will not update it, if the `body_cache_threshold` is changed

config/lru.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+
+		for i := range cache.Bodies {
+			if cache.Bodies[i].SizeBytes <= 0 {
+				cache.Bodies[i].SizeBytes = calculateEmailBodySize(&cache.Bodies[i])
+			}
+		}
+
+		caches = append(caches, cache)
+	}
```
andrinoff, line 144: `loadAllEmailBodyCaches` (which this replaces) had `if os.IsNotExist(err) { return nil, nil }`. The missing guard means fresh installs or users who’ve never opened an email will see:

```bash
Failed to load LRU from disk: open /home/user/.cache/matcha/email_bodies: no such file or directory
```

config/lru.go
```diff
@@ -0,0 +1,267 @@
...
+	once.Do(
+		func() {
+			lru = &LRU{
+				threshold: threshold,
+				cache:     make(map[string]*list.Element),
+				ll:        list.New(),
+			}
+
+			if err := lru.LoadFromDisk(); err != nil {
+				fmt.Printf("Failed to load LRU from disk: %v\n", err)
```
andrinoff: we use `log.Printf`

andrinoff (changes requested):

andrinoff:
> @andrinoff 
> 
> 
> 
> Fine, I’m just busy right now. When I get back home, I’ll fix those issues and then commit the changes.

Yeah, no worries, I'm just trying to triage as many PRs as I can, i'm very busy

mavonx:
@andrinoff 

I’ll check why the tests are failing tonight, but I’ve already addressed your RCs.

mavonx:
@andrinoff 

I think we're ready now.

andrinoff (changes requested):
The `LRU` struct has no `sync.Mutex` or `sync.RWMutex`. Every method — `Get, Put, Delete, evict`, and now the threshold update in GetLRUInstance — touches `lru.cache`, `lru.ll`, and `lru.currentSize` without any lock. 

After it, i'll merge. This could be critical in case of a race

andrinoff (approved):
lgtm

PR #1283, opened 2026-05-13
feat: maildir support

### LINKED ISSUE #1140: FEAT: Local maildir import
### Describe the feature

`matcha import maildir <path>` ingests an offlineimap / mbsync / isync / mu / notmuch maildir into the local cache so users can switch from those tools without losing history.

### Why this matters

Onboarding from existing terminal mail setups is a major adoption blocker. Mutt users have decade-old maildirs they can't bring with them.

### Notes

- Walk `cur/` and `new/`, parse the message, slot into our cache schema.
- Use the maildir filename flags (`S`, `T`, `R`, `F`) to seed seen/answered/flagged state.

### LINKED ISSUE #1183: FEAT: support local mailboxes in maildir format

_No response_

Would it be possible to support local mailboxes? mutt supports this with `mutt -f Maildir`

_No response_

### Additional context

_No response_

andrinoff:
## What?

Adds a new `maildir` protocol to matcha's backend registry so accounts can point at a local Maildir tree on disk instead of an IMAP/JMAP/POP3 server. After picking the protocol in the add-account wizard and supplying a path (e.g. `~/Mail` or `/var/mail/user`), the inbox renders messages parsed directly from the filesystem, alongside any existing remote accounts. The implementation honors the standard Maildir layout — top-level `cur/new/tmp` is treated as INBOX, and Maildir++ dot-prefixed siblings (`.Sent`, `.Archive`, `.Drafts`, …) surface as additional folders. Reading a folder promotes anything sitting in `new/` over to `cur/` (matching mutt's semantics), marking a message as read appends the `S` flag to its filename, deletion unlinks the file, and move/archive shuffle entries between Maildir folders using the emersion/go-maildir library's atomic operations. Server-side search is implemented as a local header/body scan, which is plenty fast on a local FS. Sending email and IDLE-style push notifications are unsupported and return `backend.ErrNotSupported`, since a Maildir is just a directory — no transport, no event channel. The `Capabilities()` report reflects this so the UI can dim send actions, and a `CanArchive` bit flips on automatically when a `.Archive` folder exists in the tree.

## Why?

Closes #1183 
Closes #1140

### DIFF
--- a/backend/maildir/maildir.go
+++ b/backend/maildir/maildir.go
@@ -1,0 +1,609 @@
+// Package maildir implements the backend.Provider interface for local
+// Maildir mailboxes (the `mutt -f Maildir` style). It is read/edit only —
+// there is no SMTP transport, so SendEmail returns ErrNotSupported.
+//
+// Folder layout follows Maildir++:
+//   - The configured root path is "INBOX".
+//   - Sibling directories prefixed with "." (e.g. ".Sent", ".Archive") are
+//     additional folders. Inner dots map to a "/" hierarchy.
+package maildir
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"mime"
+	"net/mail"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+	"time"
+
+	emaildir "github.com/emersion/go-maildir"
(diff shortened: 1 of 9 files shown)


floatpanebot (approved):

PR #1284, opened 2026-05-13
fix: SMTP HELO hostname

### LINKED ISSUE #1084: BUG: SMTP HELO hostname hardcoded to 'localhost'
### Describe the bug

`sender/sender.go:666` and `sender/sender.go:862`:

```go
if err = c.Hello("localhost"); err != nil { ... }
```

Several anti-spam checks reject `HELO localhost` from clients on a public network. Reputable IPs still get reliability degradation.

### Expected behavior

Use `os.Hostname()` (with a fallback) so the HELO matches the real client hostname.

nanookclaw:
## What?

Fixes #1084 by using the local OS hostname for the SMTP `HELO`/`EHLO` greeting instead of always sending `localhost`.

Both normal message sending and calendar replies now call a shared helper that returns `os.Hostname()` when available, with `localhost` kept as the fallback for hostname lookup failures or empty hostnames.

## Why?

Some SMTP anti-spam checks reject or penalize public clients that identify as `localhost`. Sending the real client hostname matches the issue's requested behavior while preserving a safe fallback if hostname lookup is unavailable.

Verification: `git diff --check` passed. `go test ./sender` could not complete locally because this environment is missing `libpcsclite` (`Package 'libpcsclite', required by 'virtual:world', not found`); the focused hostname helper test was added but the package build reaches the existing smart-card dependency first.

### DIFF
--- a/sender/sender.go
+++ b/sender/sender.go
@@ -80,6 +80,7 @@ func (a *loginAuth) Next(fromServer []byte, more bool) ([]byte, error) {
 // variable so tests can swap it with a deterministic or failing reader. By
 // default it is crypto/rand.Reader.
 var randReader io.Reader = rand.Reader
+var osHostname = os.Hostname
 
 // smimeOuterBoundary returns a fresh, high-entropy MIME boundary for an S/MIME
 // multipart/signed wrapper. If crypto/rand cannot supply randomness it returns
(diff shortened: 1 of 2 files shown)


floatpanebot (dismissed):
Hi @nanookclaw! Please fix the following issues with your PR:

sender/sender.go
```diff
@@ -92,7 +93,14 @@ func smimeOuterBoundary() (string, error) {
 	return "signed-" + fmt.Sprintf("%x", rb[:]), nil
 }
 
-// generateMessageID creates a unique Message-ID header.
```
andrinoff, line 95: why is the comment deleted?

andrinoff (dismissed):
return the comment back

nanookclaw:
Restored the function comment that was accidentally displaced and added a short comment for the new HELO helper in `965b1ba`.

Verification rerun:

```bash
git diff --check
go test ./sender
```

`git diff --check` passes. `go test ./sender` still stops before package build in this environment because `libpcsclite` is unavailable via pkg-config, matching the original local blocker.

andrinoff (approved):
lgtm

PR #1290, opened 2026-05-14
fix(clib): escape markdown fallback

### LINKED ISSUE #571: FEAT: Better fallback for MarkdownToHTML failure
In `clib/markdown.go`, if the C function `md4c_to_html` fails, it returns `nil`, and the Go wrapper `MarkdownToHTML` simply returns the original markdown bytes as-is.

If conversion fails, it would be better to:
1. Log an error (at least for debugging).
2. Perhaps provide a minimal HTML wrapper around the plain text so it's still somewhat readable in the HTML viewer.

Returning the original bytes (current behavior), which causes the HTML viewer to try and parse raw Markdown as HTML, leading to poor rendering.

### Additional context
```go
	result := C.md4c_to_html((*C.char)(cInput), C.size_t(len(md)), &outLen)
	if result == nil {
		return md // fallback to original on failure
	}
```

FromSi:
## What?

Adds a safe HTML fallback for Markdown conversion failures. When `MarkdownToHTML` fails in either the cgo `md4c` path or the no-cgo `goldmark` path, it now logs the failure and returns escaped plain text wrapped in `<pre>` instead of returning raw Markdown bytes.

Also adds a focused test for the fallback escaping behavior.

image

## Why?

Closes #571

Previously, failed Markdown conversion returned the original Markdown bytes. The email viewer then tried to parse those raw bytes as HTML, which could render poorly and treat HTML-looking plain text as markup.

The new fallback keeps failed conversions readable in the HTML viewer while escaping unsafe HTML-like input.

image

### DIFF
--- a/clib/markdown.go
+++ b/clib/markdown.go
@@ -56,7 +56,10 @@ static char* md4c_to_html(const char* input, size_t input_len, size_t* out_len) 
 }
 */
 import "C"
-import "unsafe"
+import (
+	"log"
+	"unsafe"
+)
 
 // MarkdownToHTML converts Markdown bytes to HTML using md4c (C).
 // This is significantly faster than goldmark for large documents.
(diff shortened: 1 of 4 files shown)


andrinoff (approved):
great! thanks, lgtm

PR #1292, opened 2026-05-14
fix(pop3): batch delete in one session

### LINKED ISSUE #555: BUG: POP3 batch delete is inefficient
### Describe the bug
The POP3 provider's `DeleteEmails` implementation calls `DeleteEmail` in a loop. Each call to `DeleteEmail` creates a new connection, sends a `DELE` command, and then `QUIT`s to commit. This is extremely inefficient for batch deletions.

### To reproduce
1. Use a POP3 account.
2. Select multiple emails and delete them.
3. Observe multiple connection/authentication cycles.

### Expected behavior
`DeleteEmails` should create a single connection, send all `DELE` commands, and then `QUIT` once.

### Additional context
Relevant code in `backend/pop3/pop3.go`.

FromSi:
## What?

POP3 batch deletion now sends requests like this:

```text
CONNECT
AUTH
UIDL
DELE 1
DELE 2
DELE 3
QUIT
```

`QUIT` is sent once at the end, so all `DELE` commands are committed in one POP3 session.

## Why?

Closes #555

Previously, batch deletion sent requests like this:

```text
CONNECT
AUTH
UIDL
DELE 1
QUIT

CONNECT
AUTH
UIDL
DELE 2
QUIT

CONNECT
AUTH
UIDL
DELE 3
QUIT
```

So deleting multiple emails repeated the full connection/authentication cycle for every message.

### DIFF
--- a/backend/pop3/pop3.go
+++ b/backend/pop3/pop3.go
@@ -219,12 +219,10 @@ func (p *Provider) DeleteEmails(_ context.Context, _ string, uids []uint32) erro
 	for _, uid := range uids {
 		msgID, ok := messageIDsByUID[uid]
 		if !ok {
-			conn.Quit()
 			return fmt.Errorf("pop3: message with UID %d not found", uid)
 		}
 
 		if err := conn.Dele(msgID); err != nil {
-			conn.Quit()
 			return fmt.Errorf("pop3 dele: %w", err)
 		}
 	}
--- a/backend/pop3/pop3.go
+++ b/backend/pop3/pop3.go
@@ -188,25 +188,8 @@ func (p *Provider) MarkAsUnread(_ context.Context, _ string, _ uint32) error {
 	return nil
 }
 
-func (p *Provider) DeleteEmail(_ context.Context, _ string, uid uint32) error {
-	conn, err := p.connect()
-	if err != nil {
-		return err
-	}
-
-	msgID, err := p.findMessageByUID(conn, uid)
-	if err != nil {
-		conn.Quit()
-		return err
-	}
-
-	if err := conn.Dele(msgID); err != nil {
-		conn.Quit()
-		return fmt.Errorf("pop3 dele: %w", err)
-	}
-
-	// Quit commits the deletion
-	return conn.Quit()
+func (p *Provider) DeleteEmail(ctx context.Context, folder string, uid uint32) error {
+	return p.DeleteEmails(ctx, folder, []uint32{uid})
 }
 
 func (p *Provider) ArchiveEmail(_ context.Context, _ string, _ uint32) error {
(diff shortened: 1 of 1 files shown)


FromSi:
@andrinoff

One related thing I noticed while testing locally: I could not get messages to load through the POP3 backend path itself. Messages only appeared for me when the account config also had IMAP settings with `"service_provider": "custom"`.

That creates a mismatch for this case: the message IDs shown in the UI come from IMAP, but batch deletion is then executed through the POP3 provider. POP3 maps messages by hashed UIDL values, so the IMAP UIDs do not match the POP3 UID lookup and batch deletion can fail with `pop3: message with UID ... not found`.

This might be an issue with my local setup/configuration, but I think it would be good for someone else to verify this case with a real POP3 account.

andrinoff:
I'll check it out, thank you, @FromSi

backend/pop3/pop3.go
```diff
@@ -212,14 +195,36 @@ func (p *Provider) MoveEmail(_ context.Context, _ uint32, _, _ string) error {
...
-		if err := p.DeleteEmail(ctx, folder, uid); err != nil {
-			return err
+		msgID, ok := messageIDsByUID[uid]
+		if !ok {
+			conn.Quit()
+			return fmt.Errorf("pop3: message with UID %d not found", uid)
+		}
+
+		if err := conn.Dele(msgID); err != nil {
+			conn.Quit()
```
andrinoff: If `Dele` fails on the third UID after two successful `DELE` commands, `conn.Quit()` commits the first two deletions and the function returns an error. The caller sees an error and assumes nothing was deleted — but two messages are already gone.

andrinoff (dismissed):
i will check out the behaivor you commented on a bit later, this is the only issue i see with this PR

backend/pop3/pop3.go
```diff
@@ -212,14 +195,36 @@ func (p *Provider) MoveEmail(_ context.Context, _ uint32, _, _ string) error {
...
+		conn.Quit()
+		return err
+	}
+
 	for _, uid := range uids {
-		if err := p.DeleteEmail(ctx, folder, uid); err != nil {
-			return err
+		msgID, ok := messageIDsByUID[uid]
+		if !ok {
+			conn.Quit()
```
andrinoff: same here, btw. on any error path, close the connection without `QUIT`. [RFC 1939](https://www.ietf.org/rfc/rfc1939.txt) guarantees the server will discard all `DELE` marks if the session ends without a proper `QUIT`.

FromSi: @andrinoff So the fix is just to remove `conn.Quit()` and keep only `return fmt.Errorf(...)`?

andrinoff: Yea, because an error will cause a surface of the error, but it won't say "2 deleted, third not", but even if it did, it's better to not delete any emails, if 1 of them fails

FromSi: @andrinoff updated, though I won't be able to test it :(

andrinoff: i'll check it out a bit later. Your fix is absolutely correct, as far as the old code was.

andrinoff: @FromSi also sent you an email, btw.

andrinoff:
@FromSi make your membership to the org public, please. Also, if you can -- join the [Discord](https://discord.gg/RxNrJgfatk), so that it is easier to talk to you.

andrinoff:
resolved conflicts

andrinoff (approved):
lgtm

PR #1294, opened 2026-05-15
ci: correct nix pr bump

andrinoff:
## What?

Fixes vendor hash problem

## Why?

Workflow failing


floatpanebot (approved):

PR #1299, opened 2026-05-17
feat(settings): add password meter

### LINKED ISSUE #628: FEAT: Add real-time password match indicator in settings
Is your feature request related to a problem? Password confirmation field shows no feedback until submit. Users cannot see if passwords match while typing. No strength indicator.

Describe the solution you would like: Show inline feedback:
- "✓ Passwords match" in green when they match
- "✗ Passwords do not match" in red when different
- Optional: Add strength meter (weak/medium/strong)

Describe alternatives you have considered: Only error on submit (current). Real-time feedback prevents errors.

Additional context:
- File: tui/settings.go (lines 800-850)
- UX issue: No inline validation feedback
- Complexity: Easy-Medium - compare fields on change

FromSi:
## What?

Adds real-time password feedback to the encryption settings screen.

- Shows whether the password and confirmation fields match while typing.
- Shows an informational password strength level: `weak`, `medium`, or `strong`.
- Adds `internal/passwordstrength` with a `Meter` interface and a `LibMeter` implementation backed by `github.com/wagslane/go-password-validator`.

## Why?

Closes #628

Users previously only learned that password confirmation failed after submitting the form. Inline feedback makes the encryption setup flow clearer and helps users catch mistakes before attempting to enable encryption.

The strength meter is informational only: weak passwords are not rejected.

Demo password used in the attached GIF: `CorrectHorseBatteryStaple1!`

detailed_dates


andrinoff (approved):
lgtm

PR #1300, opened 2026-05-17
ci: nixOS special version

andrinoff:
## What?

Separated `go` and `toolchain` versions, makes workflow dispatch only, adds a workflow with a cron job to check nixos go version

## Why?

@r-ryantm fails due to `go 1.26.3` being non-existent on the master branch of nixos/nix-pkgs. This will make it so that: if the user/workflow can use the latest version of Go --> it downloads and uses it, otherwise it uses the version set in `go` (which will be up to date to nixos-pkgs master version).


floatpanebot (approved):

andriy-chernov (approved):
lgtm, hope this works 🫢

PR #1304, opened 2026-05-17
fix: add TLS session resumption

### LINKED ISSUE #1174: BUG: TLS sessions are not resumed across reconnects
### Describe the bug

Every IMAP/SMTP reconnect performs a full TLS handshake. There's no `tls.Config.ClientSessionCache` configured, so server-issued session tickets are dropped. Reconnect cost on slow networks (or post-IDLE-timeout) is multiple seconds when it could be a single round-trip.

### Expected behavior

Configure a `tls.LRUClientSessionCache` per account (or shared). Verify session resumption is happening with a debug log.

### Why it's hard

Both IMAP (via `connectWithOptions`) and SMTP (`tls.Dial`) bypass the cache today. The fix is small but has to be applied in three places consistently and the cache lifetime must be tied to account credentials.

mavonx:
## What?

Configured `TLS` session resumption for `IMAP` and `SMTP`.

## Why?

Reduce reconnect latency by avoiding full `TLS` handshakes.

Closes #1174

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -232,6 +232,13 @@ func (a *Account) GetSMTPServer() string {
 	}
 }
 
+func (a *Account) GetClientSessionCache() tls.ClientSessionCache {
+	if a.ClientSessionCache == nil {
+		a.ClientSessionCache = tls.NewLRUClientSessionCache(64)
+	}
+	return a.ClientSessionCache
+}
+
 // GetSMTPPort returns the SMTP port for the account.
 func (a *Account) GetSMTPPort() int {
 	switch a.ServiceProvider {
@@ -634,7 +641,6 @@ func LoadConfig() (*Config, error) {
 			POP3Server:         rawAcc.POP3Server,
 			POP3Port:           rawAcc.POP3Port,
 			CatchAll:           rawAcc.CatchAll,
-			ClientSessionCache: tls.NewLRUClientSessionCache(64),
 		}
 
 		// Validate PGPKeySource
--- a/config/config.go
+++ b/config/config.go
@@ -31,6 +31,11 @@ const (
 	DateFormatEU  = "DD/MM/YYYY HH:MM"
 )
 
+type SessionCache struct {
+	once  sync.Once
+	cache tls.ClientSessionCache
+}
+
 // Account stores the configuration for a single email account.
 type Account struct {
 	ID              string `json:"id"`
@@ -48,8 +53,7 @@ type Account struct {
 	// regardless of which address they were delivered to.
 	CatchAll bool `json:"catch_all,omitempty"`
 
-	ClientSessionCache tls.ClientSessionCache `json:"-"` // "-" prevents the ClientSessionCache from being saved to config.json
-	sessionCacheOnce   sync.Once              `json:"-"` // "-" prevents the sessionCacheOnce from being saved to config.json
+	SC *SessionCache `json:"-"` // "-" prevents the SessionCache from being saved to config.json
 
 	// Custom server settings (used when ServiceProvider is "custom")
 	IMAPServer string `json:"imap_server,omitempty"`
@@ -235,11 +239,11 @@ func (a *Account) GetSMTPServer() string {
 }
 
 func (a *Account) GetClientSessionCache() tls.ClientSessionCache {
-	a.sessionCacheOnce.Do(func() {
-		a.ClientSessionCache = tls.NewLRUClientSessionCache(64)
+	a.SC.once.Do(func() {
+		a.SC.cache = tls.NewLRUClientSessionCache(64)
 	})
 
-	return a.ClientSessionCache
+	return a.SC.cache
 }
 
 // GetSMTPPort returns the SMTP port for the account.
@@ -593,6 +597,7 @@ func LoadConfig() (*Config, error) {
 						Password:        legacyConfig.Password,
 						ServiceProvider: legacyConfig.ServiceProvider,
 						FetchEmail:      legacyConfig.Email,
+						SC:              &SessionCache{},
 					},
 				},
 			}
@@ -644,6 +649,7 @@ func LoadConfig() (*Config, error) {
 			POP3Server:         rawAcc.POP3Server,
 			POP3Port:           rawAcc.POP3Port,
 			CatchAll:           rawAcc.CatchAll,
+			SC:                 &SessionCache{},
 		}
 
 		// Validate PGPKeySource
--- a/config/config.go
+++ b/config/config.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/google/uuid"
 	"github.com/zalando/go-keyring"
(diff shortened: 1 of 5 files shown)


fetcher/fetcher.go
```diff
@@ -371,11 +371,18 @@ func connectWithOptions(account *config.Account, extraOpts *imapclient.Options)
...
+	account.EnsureSessionCache()
+
 	options := &imapclient.Options{
 		TLSConfig: &tls.Config{
 			ServerName:         imapServer,
 			InsecureSkipVerify: account.Insecure,
 			MinVersion:         tls.VersionTLS12,
+			ClientSessionCache: account.ClientSessionCache,
+			VerifyConnection: func(cs tls.ConnectionState) error {
+				log.Printf("SMTP connection resumed: %t", cs.DidResume)
```
FromSi: ?
```go
log.Printf("IMAP TLS session resumed: %t", cs.DidResume)
```

config/config.go
```diff
@@ -244,6 +247,12 @@ func (a *Account) GetSMTPPort() int {
 	}
 }
 
+func (a *Account) EnsureSessionCache() {
+	if a.ClientSessionCache == nil {
+		a.ClientSessionCache = tls.NewLRUClientSessionCache(64)
+	}
+}
```
FromSi: If the same account is used concurrently, for example by a background IMAP reconnect and an SMTP send, this can cause a data race during the first cache initialization. It would be better to initialize the cache up front when loading/creating the account, or protect it with a mutex / `sync.Once`.

FromSi (changes requested):

FromSi:
@andrinoff Need to verify the expected behavior: 
> Verify session resumption is happening with a debug log.

I'm not entirely sure if this is implemented correctly or not.

mavonx:
> @andrinoff Need to verify the expected behavior:
> 
> > Verify session resumption is happening with a debug log.
> 
> I'm not entirely sure if this is implemented correctly or not.

@andrinoff @FromSi

I've verified the behavior in both scenarios:

- **Scenario 1**
  - Log in with your account (no `config.json` yet).
  - Click on `View Inbox`. A new IMAP connection is established because there is no `folder_cache.json`. At this stage, I got `cs.DidResume = false`.
  - I fetched a new email body that was not present in the `email_bodies` folder, so a new IMAP connection was established again, but this time with `cs.DidResume = true`.
- **Scenario 2**
  - Same as the first scenario, but this time we test with an already existing email, so the app reads directly from `config.json`.

andrinoff:
i will check it out tomorrow evening, im busy

mavonx:
I tried to handle the race condition as @FromSi suggested above, but tests failed with:

> github.com/floatpane/matcha/config.Account contains sync.Once contains sync.noCopy

I moved to a pointer-based approach by extracting `sync.Once` and `tls.ClientSessionCache` into a separate `SessionCache` struct, but this introduced a new failure in `TestSaveAndLoadConfig` — `expectedConfig` creates accounts without `SC`, while `LoadConfig` now initializes them with `SC: &SessionCache{}` so `reflect.DeepEqual` sees a mismatch between `nil` and a non-nil pointer.

As a temporary fix I added `SC: &SessionCache{}` to the accounts in `expectedConfig` to match the loaded config, and the test passes. But I'd prefer not to touch the test.

Is there a cleaner approach here?

fetcher/fetcher.go
```diff
@@ -376,6 +376,11 @@ func connectWithOptions(account *config.Account, extraOpts *imapclient.Options)
 			ServerName:         imapServer,
 			InsecureSkipVerify: account.Insecure,
 			MinVersion:         tls.VersionTLS12,
+			ClientSessionCache: account.GetClientSessionCache(),
+			VerifyConnection: func(cs tls.ConnectionState) error {
+				log.Printf("IMAP TLS connection resumed: %t", cs.DidResume)
+				return nil
+			},
```
andrinoff, line 384: this will fire on every connection

sender/sender.go
```diff
@@ -660,6 +661,11 @@ func SendEmail(account *config.Account, to, cc, bcc []string, subject, plainBody
 		ServerName:         smtpServer,
 		InsecureSkipVerify: account.Insecure,
 		MinVersion:         tls.VersionTLS12,
+		ClientSessionCache: account.GetClientSessionCache(),
+		VerifyConnection: func(cs tls.ConnectionState) error {
+			log.Printf("SMTP TLS connection resumed: %t", cs.DidResume)
+			return nil
+		},
```
andrinoff, line 668: this will too

sender/sender.go
```diff
@@ -870,10 +876,16 @@ func SendCalendarReply(account *config.Account, to []string, subject, plainBody
...
+
 	tlsConfig := &tls.Config{
 		ServerName:         smtpServer,
 		InsecureSkipVerify: account.Insecure,
 		MinVersion:         tls.VersionTLS12,
+		ClientSessionCache: account.GetClientSessionCache(),
+		VerifyConnection: func(cs tls.ConnectionState) error {
+			log.Printf("SMTP TLS connection resumed: %t", cs.DidResume)
+			return nil
+		},
```
andrinoff, line 888: and here

config/config.go
```diff
@@ -229,6 +238,14 @@ func (a *Account) GetSMTPServer() string {
...
 }
 
+func (a *Account) GetClientSessionCache() tls.ClientSessionCache {
+	a.SC.once.Do(func() {
+		a.SC.cache = tls.NewLRUClientSessionCache(64)
+	})
+
+	return a.SC.cache
+}
+
```
andrinoff, line 248: `GetClientSessionCache()` panics if `SC` is `nil`. Won't happen right now, but better add a nil guard. better be safe, than sorry

andrinoff (changes requested):

andrinoff:
@mavonx fix the debug logging. 

image

this should not be the experience

andrinoff:
debug logging could be done by having an env variable (set with `ENV_VAR=1 matcha`) and it will then spit out.

andrinoff:
actually, use #1311, just merged, and rebased the branch

mavonx:
> `GetClientSessionCache()` panics if `SC` is `nil`. Won't happen right now, but better add a nil guard. better be safe, than sorry

@andrinoff 

Regarding the nil guard — adding a nil check in `GetClientSessionCache()` would introduce a race condition: two goroutines could both see `SC == nil` simultaneously and each allocate a new `&SessionCache{}`, with one overwriting the other and losing cached sessions, so instead I initialize `SC` eagerly at every account creation site (new login, `LoadConfig`, and legacy migration) to eliminate the race condition — which is what forced `TestSaveAndLoadConfig` to need `SC: &SessionCache{}` in `expectedConfig` to match what `LoadConfig` returns. 

Is that acceptable, or do you have a preferred approach to avoid touching the test?

andrinoff:
ok, that's fine then. do the log, i'll resolve the comment

andrinoff:
i fixed up this pr... why did you force-push...

mavonx:
@andrinoff 

Sorry about that! It’s my first time doing a rebase, but I think everything is clean now.

mavonx:
@andrinoff 

Sorry about that! I think everything is clean now.

andrinoff (approved):
looks good to me, @FromSi will review later (merge after your review, please)

FromSi (approved):

floatpanebot (changes requested):
The CI workflow failed. Please fix the following issues locally and push again:

Please check the [CI logs](https://github.com/floatpane/matcha/actions/runs/26239612363) for more details on the failure.

Issue #981, opened 2026-04-24
BUG: Goroutine in main.go:3620 has no panic recovery


andrinoff:
### Describe the bug

In main.go:3620, a goroutine runs macOS sync operations:
```go
go func() {
    _ = config.SyncMacOSContacts()
    _ = theme.SyncWithMacOS()
}()
```

Both errors ignored AND no panic recovery. If either function panics, app crashes silently.

### To reproduce

Run on macOS with corrupted contacts database.

### Expected behavior

Add panic recovery. Log errors instead of ignoring them.

Issue #980, opened 2026-04-24
BUG: Goroutine in main.go:908 has no panic recovery


andrinoff:
### Describe the bug

In main.go:908, a goroutine is started without panic recovery:
```go
go func() {
    // background operation
}()
```

Unrecovered panics in background goroutines crash the entire app.

### To reproduce

Trigger a panic in the background goroutine.

### Expected behavior

Add panic recovery with logging to all fire-and-forget goroutines.

floatpanebot:
This has had no activity for 30 days. It will be closed in 14 days unless updated. Comment or remove the `stale` label to keep it open.

floatpanebot:
Closing due to inactivity. Reopen if still relevant.

Issue #979, opened 2026-04-24
BUG: Goroutine in main.go:625 has no panic recovery


andrinoff:
### Describe the bug

In main.go:625, a goroutine is started for cache saving:
```go
go func() {
    // save cache operations
}()
```

If this goroutine panics (e.g., nil map access, file system error), it crashes the entire TUI application with no recovery or error reporting.

### To reproduce

1. Trigger cache save with corrupted in-memory state
2. Goroutine panics, app crashes

### Expected behavior

Add `defer func() { if r := recover(); r != nil { log.Printf(...) } }()` to goroutines that can run independently.

floatpanebot:
This has had no activity for 30 days. It will be closed in 14 days unless updated. Comment or remove the `stale` label to keep it open.

floatpanebot:
Closing due to inactivity. Reopen if still relevant.

PR #1305, opened 2026-05-18
fix(daemon): recover acceptLoop panics

### LINKED ISSUE #1119: BUG: daemon acceptLoop has no panic recovery
### Describe the bug

`daemon/daemon.go:120` launches `go d.acceptLoop()` without a recover. A panic inside a connection handler that propagates back through Accept (or a panic from go-imap at this layer) takes down the whole daemon process.

### Expected behavior

Wrap the goroutine body in a recover that logs and restarts the loop. (Same pattern as #979/#980/#981 for main.go goroutines.)

mvanhorn:
## What?

Adds a top-level `defer func() { recover(); log }()` to `(*Daemon).acceptLoop` so a panic in the accept goroutine logs with a stack trace instead of taking down the whole daemon.

Closes #1119

## Why?

`daemon/daemon.go:120` launches `go d.acceptLoop()` with no panic recovery. A panic inside the loop (or in `addClient` / `daemonrpc.NewConn`) propagates up and the runtime aborts the process. The sync loop, IDLE watcher, and connected clients all die with it. The issue notes the same risk applies to the prior recovery fixes (#979 / #980 / #981).

The recovery pattern matches the existing one at `main.go:4089`: log the panic with a `runtime/debug.Stack()` trace and let the goroutine exit. The daemon stops accepting new connections after a recovered panic, but already-connected clients and the background sync loop keep running, which is what the issue body asks for.

### Verification

- `go build ./...` clean
- `go vet ./daemon/...` clean
- `go test ./daemon/...` passes
- Diff is +6 lines in `daemon/daemon.go` (one import, one `defer` block)

### DIFF
--- a/daemon/daemon.go
+++ b/daemon/daemon.go
@@ -185,25 +185,31 @@ func (d *Daemon) initProviders() {
 }
 
 func (d *Daemon) acceptLoop() {
-	defer func() {
-		if r := recover(); r != nil {
-			log.Printf("daemon: acceptLoop panic recovered: %v\n%s", r, debug.Stack())
-		}
-	}()
 	for {
-		conn, err := d.listener.Accept()
-		if err != nil {
-			select {
-			case <-d.shutdown:
-				return
-			default:
-				log.Printf("daemon: accept error: %v", err)
-				continue
+		done := func() bool {
+			defer func() {
+				if r := recover(); r != nil {
+					log.Printf("daemon: acceptLoop panic recovered: %v\n%s", r, debug.Stack())
+				}
+			}()
+			conn, err := d.listener.Accept()
+			if err != nil {
+				select {
+				case <-d.shutdown:
+					return true
+				default:
+					log.Printf("daemon: accept error: %v", err)
+					return false
+				}
 			}
+			rpcConn := daemonrpc.NewConn(conn)
+			d.addClient(rpcConn)
+			go d.handleClient(rpcConn)
+			return false
+		}()
+		if done {
+			return
 		}
-		rpcConn := daemonrpc.NewConn(conn)
-		d.addClient(rpcConn)
-		go d.handleClient(rpcConn)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


floatpanebot (dismissed):

- **Title**: Is too long (45 characters). The PR title must be strictly under 40 characters.

daemon/daemon.go
```diff
@@ -184,6 +185,11 @@ func (d *Daemon) initProviders() {
 }
 
 func (d *Daemon) acceptLoop() {
+	defer func() {
+		if r := recover(); r != nil {
+			log.Printf("daemon: acceptLoop panic recovered: %v\n%s", r, debug.Stack())
+		}
+	}()
```
FromSi: This does not match the expected behavior: `acceptLoop` exits, while it should restart/continue running. At minimum, we could wrap the current loop body in an inner `func` inside the `for` loop, move the existing logic there, and place the `defer func()` recovery inside that inner function.

From issue:
> Wrap the goroutine body in a recover that logs and restarts the loop.

FromSi (changes requested):

daemon/daemon.go
mvanhorn: Good catch, thanks. You're right - the function-level `defer recover()` caught the panic but `acceptLoop` exited instead of restarting, which contradicts the issue's stated intent.

Fixed in 35ce26a by wrapping the loop body in an inner `func() bool` with its own `defer recover()`, exactly as you suggested. The inner func returns `done` so the existing shutdown path still exits the outer loop cleanly; the panic-recovery and accept-error paths both return false and let the for loop continue.

`go vet`, `gofmt`, and `go test ./daemon/` all pass locally.

FromSi (approved):

andrinoff (approved):

mvanhorn:
Thanks @andrinoff for landing the recover. A panic in acceptLoop taking down the whole daemon was a real footgun; logging with a stack trace and surviving is the right behavior.

PR #1306, opened 2026-05-18
fix(daemon): add panic recovery

### LINKED ISSUE #1121: BUG: handleRefreshFolder background goroutine has no panic recovery
### Describe the bug

`daemon/handler.go:297` starts a `go func()` for the async refresh without a recover. A panic from the provider (e.g. nil deref in JMAP body parse) takes down the daemon.

### Expected behavior

Wrap with recover and broadcast a SyncError to subscribers instead of crashing.

mavonx:
## What?

Added panic recovery to the async refresh goroutine in `daemon/handler.go`.

## Why?

The goroutine had no panic recovery. A panic from any provider would crash the entire daemon. Now it recovers and notifies subscribers instead.

Closes #1121

### DIFF
--- a/daemon/handler.go
+++ b/daemon/handler.go
@@ -310,6 +310,17 @@ func (d *Daemon) handleRefreshFolder(conn *daemonrpc.Conn, req *daemonrpc.Reques
 
 	// Async: fetch in background, push events when done.
 	go func() {
+
+		defer func() {
+			if r := recover(); r != nil {
+				d.broadcastToSubscribers(params.AccountID, params.Folder, daemonrpc.EventSyncError, daemonrpc.SyncErrorEvent{
+					AccountID: params.AccountID,
+					Folder:    params.Folder,
+					Error:     fmt.Sprintf("panic: %v", r),
+				})
+			}
+		}()
+
 		p, err := d.getProvider(params.AccountID)
 		if err != nil {
 			log.Printf("daemon: refresh provider error: %v", err)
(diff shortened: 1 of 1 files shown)


daemon/handler.go
```diff
@@ -310,6 +310,17 @@ func (d *Daemon) handleRefreshFolder(conn *daemonrpc.Conn, req *daemonrpc.Reques
 
 	// Async: fetch in background, push events when done.
 	go func() {
+
+		defer func() {
+			if r := recover(); r != nil {
+				d.broadcastToSubscribers(params.AccountID, params.Folder, daemonrpc.EventSyncError, daemonrpc.SyncErrorEvent{
```
FromSi, line 316: Add `log.Printf` before `broadcastToSubscribers` so the daemon records the original refresh panic even if notifying subscribers succeeds or no clients are subscribed.:

```php
log.Printf("daemon: refresh panic for account=%s folder=%s: %v", params.AccountID, params.Folder, r)
```

daemon/handler.go
```diff
@@ -310,6 +310,17 @@ func (d *Daemon) handleRefreshFolder(conn *daemonrpc.Conn, req *daemonrpc.Reques
 
 	// Async: fetch in background, push events when done.
 	go func() {
+
```
FromSi: Remove empty line after `go func() {`

FromSi (changes requested):

FromSi:
lgtm

FromSi (approved):

andrinoff (approved):

PR #1311, opened 2026-05-20
feat(cli): add log verbosity flags

### LINKED ISSUE #576: FEAT: CLI flag for log verbosity
There is no easy way to control the verbosity of logging from the CLI.

Add a `--verbose` or `-v` flag (and maybe `--debug`) to set the logging level.

N/A

### Additional context
Check how `log` package is used throughout the repo.

mvanhorn:
## What?

Adds two global CLI flags for log verbosity:

- `--verbose` / `-V` raises the log level to verbose, which emits a `matcha: loaded config with N account(s)` line once config loads. Reserved for additional sites later.
- `--debug` raises the log level to debug, which enables the two existing `debug:` log lines in the email-body cache path so they only fire when explicitly requested.

A small `internal/loglevel` package holds the level state and the `Debugf` / `Verbosef` / `Infof` helpers. Level state is `sync/atomic`-backed so concurrent goroutines stay race-clean. The two pre-existing `debug:` log sites in `main.go` are routed through `loglevel.Debugf`.

The existing `-v` / `--version` short-circuit is preserved unchanged. `--verbose` uses `-V` (capital) as its short form to avoid colliding with `--version`.

Flags are parsed before subcommand dispatch by a small `parseGlobalFlags` helper that strips them from `os.Args` so subcommand parsers don't see them.

Documented in README under a new "Logging" section.

Unit tests cover the level-gating contract: Debugf/Verbosef/Infof each fire only at their level or higher.

## Why?

Closes #576.

Today there is no way to control matcha's log verbosity from the CLI. The repo has 106 `log.Printf` call sites, two of which already include a `debug:` prefix that runs unconditionally. The issue asked for `--verbose` / `--debug` flags to gate that output. This PR is the smallest change that wires the flags through a level helper without re-classifying every existing log line - that broader cleanup can happen incrementally.

Default behavior is unchanged: 104 `log.Printf` calls still fire at the default level. Only the two existing `debug:`-prefixed lines and the new `Verbosef` config-load line are gated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,18 @@ matcha send --to alice@example.com --subject "Hello" --body "Sent by my AI agent
 
 [Learn more](https://docs.matcha.email/Features/AI_AGENTS)
 
+### Logging
+
+Matcha supports global logging verbosity flags before the main command or subcommand:
+
+```bash
+matcha --verbose              # enable verbose logging
+matcha -V daemon status       # short form for verbose logging
+matcha --debug daemon status  # enable debug logging
+```
+
+The existing `-v` and `--version` flags continue to print the Matcha version.
+
 **AI Rewrite Plugin:** Matcha includes an AI rewrite plugin that allows you to rewrite your email drafts using OpenAI, Ollama, Gemini, or Claude.
 
 [Setup Guide](https://docs.matcha.email/setup-guides/ai-rewrite)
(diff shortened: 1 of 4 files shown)


floatpanebot (dismissed):

- **Title**: Is too long (56 characters). The PR title must be strictly under 40 characters.

andrinoff (approved):

mvanhorn:
Thanks @andrinoff for getting the verbosity flags in. Having `--verbose` and `--debug` as global flags makes matcha's config-load step much less guessy to troubleshoot.

Issue #1309, opened 2026-05-19
BUG: maildir account fails to fetch with "unsupported service_provider:"


jdlawrie:
### Describe the bug

I tried adding a maildir account and it looks like it still needs an imap_server for it to work (though I only briefly searched the codebase for the error)

### To reproduce

Screenshots of usage attached.

```
~ matcha -v
matcha version 0.38.0 (ddf3a5ebb1d9fc5712164281160152f6a9d9a378) built on 2026-05-14T10:40:52Z
```

<img width="1035" height="386" alt="Image" src="https://github.com/user-attachments/assets/a775b7e6-eb44-48d3-a950-19e73f0d6a7d" />
<img width="920" height="219" alt="Image" src="https://github.com/user-attachments/assets/0d56dee2-c118-4101-8064-1223071c4129" />

### Expected behavior

It should load the maildir

### Matcha version

0.38.0

### Go version

1.26.2

### OS

macOS

### OS version

macOS 15.7.2

### Architecture

arm64 (aarch64 / Apple Silicon)

### Terminal

kitty

andrinoff:
I think this is due to #1270. I'll definitely check it out soon.

andrinoff:
@jdlawrie apparently, my fault mostly. I pushed a fix #1314. Here are [binaries for you to try](https://github.com/floatpane/matcha/pull/1314#issuecomment-4502595669). I'll publish the version, as soon as we merge this and 1 other PR!

xxxserxxx:
With the preview, it now flashes the "unsupported service provider" message and then goes to the INBOX, but no emails are shown.

[Asciinema recording](https://asciinema.org/a/XI7xwL6aQ8FnjdM8)

My config is (basically)

```
{
  "accounts": [
    {
      "id": "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx",
      "name": "Xxxxxxx X Xxxxx",
      "email": "xxx@xxxx.xxx",
      "service_provider": "",
      "fetch_email": "xxx",
      "auth_method": "password",
      "protocol": "maildir",
      "pop3_port": 995,
      "maildir_path": "/home/xxx/.mail/xxx"
    }
  ]
}
```

[...]

PR #1314, opened 2026-05-20
fix: provider, than fallback to fetcher

andrinoff:
## What?

Uses the provider, than falls back to normal fetcher.

## Why?

`fetchFoldersCmd` called `fetcher.FetchFolders(&acc)` directly. That path is IMAP-only — `connectWithOptions` rejects empty `imap_server` with "unsupported service_provider". Maildir backend has its own `FetchFolders` but was bypassed.

### DIFF
--- a/main.go
+++ b/main.go
@@ -531,7 +531,7 @@ func (m *mainModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		// Fetch folders and INBOX emails in parallel (background refresh)
 		batchCmds := []tea.Cmd{
 			m.current.Init(),
-			fetchFoldersCmd(m.config),
+			m.fetchFoldersCmd(),
 			fetchFolderEmailsCmd(m.config, "INBOX"),
 			listenForIdleUpdates(m.idleUpdates),
 		}
@@ -2819,7 +2819,8 @@ func listenForDaemonEvents(ch <-chan *daemonrpc.Event) tea.Cmd {
 
 // --- Folder-based command functions ---
 
-func fetchFoldersCmd(cfg *config.Config) tea.Cmd {
+func (m *mainModel) fetchFoldersCmd() tea.Cmd {
+	cfg := m.config
 	return func() tea.Msg {
 		if !cfg.HasAccounts() {
 			return nil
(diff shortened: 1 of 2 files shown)


floatpanebot (dismissed):

- **Title**: Is too long (43 characters). The PR title must be strictly under 40 characters.

floatpanebot (dismissed):

- **Title**: Is too long (43 characters). The PR title must be strictly under 40 characters.

FromSi:
Add Closes #1309 to description

main.go
```diff
@@ -2393,7 +2435,47 @@ func loadCachedEmails() tea.Cmd {
 	}
 }
 
-func refreshEmails(cfg *config.Config, mailbox tui.MailboxKind, counts map[string]int) tea.Cmd {
+// mailboxFolderName resolves a TUI MailboxKind to a backend folder name.
+// IMAP-specific provider quirks (Gmail's "[Gmail]/Sent Mail" etc.) are
+// preserved via ServiceProvider; non-IMAP backends use logical names.
+func mailboxFolderName(acc *config.Account, mailbox tui.MailboxKind) string {
```
FromSi, line 2441: The comment says non-IMAP backends use logical folder names, but this mapping is driven only by `ServiceProvider`. If a non-IMAP account has `ServiceProvider` set, we can pass IMAP-specific names like `[Gmail]/All Mail` into JMAP/maildir providers. Should the Gmail/Outlook/iCloud mapping be gated to IMAP accounts only, and return `Sent`/`Trash`/`Archive` for other protocols?

The problematic part is that the new function checks `ServiceProvider`, but does not check `Protocol`.

So if an account has:

```go
Protocol: "maildir"
ServiceProvider: "gmail"
```

the code may resolve:

```text
Archive -> [Gmail]/All Mail
```

But for maildir it should resolve to:

```text
Archive -> Archive
```

FromSi (changes requested):

main.go
andrinoff: i honestly don't really know how else to write it. the person creating the issue does not need a provider, so it will not even go there. i'll think of it in a latter PR.

PR #1316, opened 2026-05-21
fix(tui): log image debug

### LINKED ISSUE #701: BUG: Hardcoded fmt.Print in HTML viewer
### Describe the bug
In `view/html.go:253`, `fmt.Print(msg)` is used directly for logging image protocol debug information instead of the standard logger. This bypasses log formatting and destinations.

### To reproduce
Enable image protocol debugging and observe stdout.

### Expected behavior
It should use the application's standard logging facility (e.g., `log.Printf` or a structured logger).

resolvicomai:
## What?
- route image protocol debug messages through the standard logger instead of writing directly to stdout
- add coverage that verifies debug output goes to the logger and leaves stdout untouched

## Why?
Debug output should respect the application logging path instead of bypassing logger formatting and destinations.

Fixes #701

## Tests
- go test ./view -count=1
- make lint
- go test ./...

### DIFF
--- a/view/html_test.go
+++ b/view/html_test.go
@@ -1,7 +1,10 @@
 package view
 
 import (
+	"bytes"
 	"fmt"
+	"io"
+	"log"
 	"os"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


floatpanebot (dismissed):

- **Title**: Is too long (41 characters). The PR title must be strictly under 40 characters.

view/html_test.go
```diff
@@ -68,6 +71,56 @@ func TestDecodeQuotedPrintable(t *testing.T) {
 	}
 }
 
+func TestDebugImageProtocolUsesLogger(t *testing.T) {
```
FromSi, line 73: I wonder if this test is more complexity than we need for this fix. It mutates global logger state and `os.Stdout` to cover a one-line logging change. Since this is not a broader logging abstraction, maybe the implementation change is enough, or the test could just assert logger output without capturing stdout.

FromSi (dismissed):
lgtm

andrinoff (changes requested):
@resolvicomai we are migrating to a new log introduced in #1311. Please use `loglevels` internal package

resolvicomai:
Thanks. I pushed a follow-up that uses the internal `loglevel` package for the image protocol debug log and simplifies the test so it only asserts logger output.

Validation:
- `go test ./view`
- `go test ./...`
- `make lint`
- `git diff --check`

FromSi (approved):

andrinoff (approved):

PR #1317, opened 2026-05-21
fix(tui): log debug file errors

### LINKED ISSUE #698: BUG: Unchecked file write error in Kitty protocol debugger
### Describe the bug
In `view/html.go:261`, `_, _ = f.WriteString(msg)` is ignored for `DEBUG_KITTY_LOG`.

### To reproduce
Enable `DEBUG_KITTY_LOG` and restrict permissions on the log file.

### Expected behavior
Error should be checked and logged.

### LINKED ISSUE #699: BUG: Unchecked file close error in Kitty protocol debugger
### Describe the bug
In `view/html.go:262`, `_ = f.Close()` ignores the close error.

### To reproduce
Enable `DEBUG_KITTY_LOG`.

### Expected behavior
Error should be handled properly.

resolvicomai:
## What?
- Route debug image protocol log writes through a helper that checks write and close errors.
- Log failures for DEBUG_IMAGE_PROTOCOL_LOG and DEBUG_KITTY_LOG instead of silently ignoring them.
- Add regression coverage for write and close failure paths.

Closes #698
Closes #699

## Why?
The Kitty/image protocol debugger previously ignored both WriteString and Close errors when writing debug logs, which could hide failed diagnostics and resource cleanup problems.

## Tests
- go test ./view -count=1 -run TestDebugImageProtocolReportsLog
- go test ./view -count=1
- go test ./...
- make lint

### DIFF
--- a/view/html.go
+++ b/view/html.go
@@ -4,7 +4,6 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io"
-	"log"
 	"mime/quotedprintable"
 	"os"
 	"regexp"
@@ -15,6 +14,7 @@ import (
 	"charm.land/lipgloss/v2"
 	"github.com/floatpane/matcha/clib"
 	"github.com/floatpane/matcha/internal/httpclient"
+	"github.com/floatpane/matcha/internal/loglevel"
 	"github.com/floatpane/matcha/theme"
 	lru "github.com/hashicorp/golang-lru/v2"
 )
@@ -278,7 +278,7 @@ var (
 	debugImageProtocolOpenLogFile = func(path string) (debugImageProtocolLogFile, error) {
 		return os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
 	}
-	debugImageProtocolLogErrorf = log.Printf
+	debugImageProtocolLogErrorf = loglevel.Infof
 )
 
 func debugImageProtocol(format string, args ...interface{}) {
@@ -286,7 +286,7 @@ func debugImageProtocol(format string, args ...interface{}) {
 		return
 	}
 	msg := fmt.Sprintf("[img-protocol] "+format+"\n", args...)
-	fmt.Print(msg)
+	loglevel.Infof("%s", strings.TrimSuffix(msg, "\n"))
 	if path := os.Getenv("DEBUG_IMAGE_PROTOCOL_LOG"); path != "" {
 		writeDebugImageProtocolLog(path, msg)
 	} else if path := os.Getenv("DEBUG_KITTY_LOG"); path != "" {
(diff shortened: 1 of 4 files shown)


floatpanebot (dismissed):

view/html.go
```diff
@@ -268,22 +269,41 @@ func imageProtocolSupported() bool {
...
-			_ = f.Close()
-		}
+		writeDebugImageProtocolLog(path, msg)
+	}
+}
+
+func writeDebugImageProtocolLog(path, msg string) {
+	f, err := debugImageProtocolOpenLogFile(path)
+	if err != nil {
+		return
```
FromSi, line 301: Should we log the `OpenFile` error here as well? The PR description says failures for `DEBUG_IMAGE_PROTOCOL_LOG` / `DEBUG_KITTY_LOG` are logged, but this path still silently returns on open failure. This also seems relevant to #698's reproduction case with restricted log-file permissions, where the failure may happen before `WriteString`.

view/html.go
```diff
@@ -268,22 +269,41 @@ func imageProtocolSupported() bool {
 		weztermSupported() || waystSupported() || warpSupported() || konsoleSupported()
 }
 
+type debugImageProtocolLogFile interface {
+	WriteString(string) (int, error)
+	Close() error
+}
+
+var (
+	debugImageProtocolOpenLogFile = func(path string) (debugImageProtocolLogFile, error) {
```
FromSi, line 278: This helper fixes the `view/html.go` debug-log path, but `fetcher/fetcher.go` has another `DEBUG_KITTY_LOG` writer that still ignores open/write/close failures. If this PR is intended to cover `DEBUG_KITTY_LOG` failures generally, could we route that path through the same checked behavior too, or narrow the PR description to the `view/html.go` path only?

FromSi (changes requested):

view/html.go
resolvicomai: Fixed in the follow-up commit: `writeDebugImageProtocolLog` now logs open failures as well as write/close failures, with regression coverage.

view/html.go
resolvicomai: Fixed in the follow-up commit: the `fetcher/fetcher.go` `DEBUG_KITTY_LOG` path now uses checked open/write/close handling too, with a focused regression test.

FromSi (dismissed):
lgtm

andrinoff (changes requested):
@resolvicomai we are migrating to a new log introduced in #1311. Please use `loglevels` internal package

resolvicomai:
Thanks. I pushed a follow-up that routes both debug-log paths through the internal `loglevel` package:

- image protocol debug output and file-error logging in `view/html.go`
- kitty body-selection output and file-error logging in `fetcher/fetcher.go`

Validation:
- `go test ./view ./fetcher`
- `go test ./...`
- `make lint`
- `git diff --check`

FromSi (approved):

andrinoff (approved):

andrinoff:
conflicts @resolvicomai

PR #1319, opened 2026-05-21
fix(i18n): wrap parse error cause

### LINKED ISSUE #1050: BUG: fmt.Errorf chain '%w: %v' loses original cause in i18n parser
### Describe the bug

In i18n/parser.go:18:
```go
return nil, fmt.Errorf("%w: %v", ErrParseFailed, err)
```

Only `ErrParseFailed` is wrapped — the underlying `err` is rendered as text and lost from the error chain. `errors.Is(returned, originalErr)` no longer matches.

### Expected behavior

Use `%w` for both: `fmt.Errorf("%w: %w", ErrParseFailed, err)` (Go ≥ 1.20) or wrap the original error directly.

resolvicomai:
## What?
- Wrap the JSON parser source error with `%w` instead of rendering it with `%v`.
- Add regression coverage that preserves both `ErrParseFailed` and the original `json.SyntaxError` in the error chain.

Closes #1050

## Why?
The i18n parser already wraps `ErrParseFailed`, but the underlying JSON parse error was converted to text. That made `errors.As` unable to recover the original cause from callers that need to inspect it.

## Tests
- go test ./i18n -count=1 -run TestParseJSONWrapsSyntaxError
- go test ./i18n -count=1
- go test ./...
- make lint
- git diff --check

### DIFF
--- a/i18n/parser.go
+++ b/i18n/parser.go
@@ -15,7 +15,7 @@ type TranslationFile struct {
 func ParseJSON(data []byte) (MessageMap, error) {
 	var file TranslationFile
 	if err := json.Unmarshal(data, &file); err != nil {
-		return nil, fmt.Errorf("%w: %v", ErrParseFailed, err)
+		return nil, fmt.Errorf("%w: %w", ErrParseFailed, err)
 	}
 
 	messages := make(MessageMap)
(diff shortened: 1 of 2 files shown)


FromSi (approved):
lgtm

andrinoff (approved):

PR #1321, opened 2026-05-21
fix: keep empty part path

### LINKED ISSUE #1113: BUG: formatPartPath silently maps empty path to '1'
### Describe the bug

`fetcher/fetcher.go:224-227`:

```go
func formatPartPath(path []int) string {
    if len(path) == 0 {
        return "1"
    }
    ...
```

Returning `"1"` for an empty path masks bugs in the caller (a Walk that produced no path is almost certainly wrong) and conflicts with parsePartID, which treats `""` and `"TEXT"` as nil and `"1"` as a real path of length 1.

### Expected behavior

Return `""` (matching parsePartID's nil-path convention) and let the caller decide.

resolvicomai:
## What?

Fixes #1113. `formatPartPath` now returns an empty string for nil/empty paths instead of silently formatting them as part `1`. A regression test covers both nil and empty slices.

## Why?

`parsePartID` already treats an empty string as the nil-path convention, while `1` is a real MIME part. Returning `1` for an empty walk path can mask caller bugs and fetch the wrong part.

Validation:
- `go test ./fetcher -run TestFormatPartPathEmptyPath -count=1`
- `go test ./fetcher -count=1`
- `go test ./...`
- `make lint`
- `git diff --check`

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -323,7 +323,7 @@ func parsePartID(partID string) []int {
 // formatPartPath converts a Walk path like []int{1, 2, 3} to "1.2.3".
 func formatPartPath(path []int) string {
 	if len(path) == 0 {
-		return "1"
+		return ""
 	}
 	parts := make([]string, len(path))
 	for i, p := range path {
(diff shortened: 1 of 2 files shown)


floatpanebot (dismissed):

FromSi (approved):
lgtm

andrinoff (approved):

PR #1318, opened 2026-05-21
fix(sender): handle qp errors

resolvicomai:
## What?
- Add a shared quoted-printable writer helper that returns write and close errors.
- Use it for plaintext-only and multipart text/html message parts.
- Add regression coverage for a failing quoted-printable flush.

Closes #614

## Why?
Ignoring `quotedprintable.Writer.Close()` can let email composition continue with an incomplete or corrupted encoded body.

## Tests
- `go test ./sender -run TestWriteQuotedPrintablePropagatesFlushError -count=1`
- `go test ./sender -count=1`
- `go test ./view -count=1`
- `go test ./...`
- `make lint`
- `git diff --check`

PR #1322, opened 2026-05-21
fix: remove flowed claim

### LINKED ISSUE #1114: BUG: Plaintext part always asserts format=flowed without space-stuffing
### Describe the bug

`sender/sender.go:225`:

```go
fmt.Fprintf(&partBuf, "Content-Type: text/plain; charset=UTF-8; format=flowed\r\n")
```

We claim format=flowed but never apply RFC 3676 space-stuffing or trailing-space soft-break encoding to the body. Conformant readers rewrap the lines and quoted text loses its `>` markers.

### Expected behavior

Either drop `format=flowed` (safer) or add a proper space-stuffer (`encodeFormatFlowed`).

resolvicomai:
## What?

Fixes #1114. Plaintext-only messages now emit `Content-Type: text/plain; charset=UTF-8` without `format=flowed`, and both plaintext MIME paths share the same header writer. A regression test covers that the header no longer advertises flowed text while keeping quoted-printable encoding.

## Why?

The sender quoted-printable encodes the body but does not implement RFC 3676 format=flowed space-stuffing or soft line break semantics. Advertising `format=flowed` can make conforming clients reflow plain text incorrectly, especially quoted text.

Validation:
- `go test ./sender -run TestWritePlainTextPartHeadersDoesNotClaimFormatFlowed -count=1`
- `go test ./sender -count=1`
- `go test ./...`
- `make lint`
- `git diff --check`

### DIFF
--- a/sender/sender.go
+++ b/sender/sender.go
@@ -195,6 +195,11 @@ func writeQuotedPrintable(w io.Writer, body string) error {
 	return nil
 }
 
+func writePlainTextPartHeaders(w io.Writer) {
+	fmt.Fprintf(w, "Content-Type: text/plain; charset=UTF-8\r\n")
+	fmt.Fprintf(w, "Content-Transfer-Encoding: quoted-printable\r\n\r\n")
+}
+
 // SendEmail constructs a multipart message with plain text, HTML, embedded images, and attachments.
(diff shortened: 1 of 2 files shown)


FromSi (dismissed):
lgtm

andrinoff:
conflicts with your own PR @resolvicomai

resolvicomai:
Thanks for flagging it. I rebased this PR on current `master` after #1318 merged and resolved the conflict by keeping the `writeQuotedPrintable` helper from #1318 plus the plaintext header helper from this PR.

Validation after rebase:
- `go test ./sender`
- `go test ./...`
- `make lint`
- `git diff --check origin/master...HEAD`

The branch is pushed again as `ac4938f`.

PR #1329, opened 2026-05-21
feat(tui): add log panel

FromSi:
## What?

Added an optional in-TUI log panel enabled with `--logs`. The panel is backed by a small in-memory logger buffer and renders the latest log entries at the bottom of the TUI without overwriting the active view.

The main TUI layout now reserves space for the log panel when it is enabled, so the current screen receives a reduced height instead of being drawn under the logs. The visual log panel itself lives in `tui/log_panel.go`.

Also updated `internal/loglevel` so logs written through it are consistently prefixed:

- `debug:`
- `verbose:`
- `info:`

Added `make run-log`, which starts Matcha with both debug logging and the in-TUI log panel enabled:

```bash
make run-log
```
view

## Why?

While reviewing recent PRs, I noticed it was hard to observe logs cleanly in the TUI. Logs written through `log.Printf` or `internal/loglevel` could appear over the Bubble Tea screen, or required redirecting stderr to a separate file.

This gives us a simple way to inspect runtime logs directly inside Matcha while testing behavior such as TLS resumption, plugin loading, background fetches, and other debug/verbose paths.

I did not search for an existing issue for this.

## Notes

This is an initial model: there is no scrolling yet, and the panel only shows the latest entries.

In the future, we can consider a mediator/multi-sink logger that writes through one logger interface to multiple targets, for example the in-memory TUI buffer, stderr, and log files.

The panel is intentionally read-only for now; keyboard handling and scrolling can be added later without changing the logger interface.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build test run clean lint fmt vet build-full generate_screenshots
+.PHONY: build test run run-log clean lint fmt vet build-full generate_screenshots
 
 BINARY_NAME=matcha
 BUILD_DIR=bin
@@ -35,6 +35,9 @@ build-full:
 
 run:
 	go run .
+
+run-log:
+	go run . --debug --logs
 
 test:
 	go test ./...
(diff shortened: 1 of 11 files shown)


andrinoff (approved):

floatpanebot (changes requested):
The CI workflow failed. Please fix the following issues locally and push again:

Please check the [CI logs](https://github.com/floatpane/matcha/actions/runs/26309704137) for more details on the failure.

PR #1337, opened 2026-05-22
fix(fetcher): close debug files

### LINKED ISSUE #715: BUG: File descriptor leak in DEBUG_IMAP logger
**Describe the bug**
In fetcher/fetcher.go lines 37-51, the DEBUG_IMAP file handle is opened via sync.Once but never closed, leaking file descriptor for application lifetime.

**To reproduce**
1. Set DEBUG_IMAP environment variable to file path
2. Start matcha
3. Examine fetcher/fetcher.go lines 37-51:
```go
var (
    debugIMAPFile *os.File
    debugIMAPOnce sync.Once
)

func getDebugIMAPWriter() io.Writer {
    debugIMAPOnce.Do(func() {
        if path := os.Getenv("DEBUG_IMAP"); path != "" {
            f, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0600)
            if err == nil {
                debugIMAPFile = f  // Never closed
            }
        }
    })
    if debugIMAPFile != nil {
        return debugIMAPFile
    }
    return nil
}
```

4. Notice file handle opened but no cleanup mechanism exists

**Expected behavior**
File descriptor should be closed on application shutdown. Options:
1. Register cleanup function via defer in main() or init()
2. Add CloseDebugFiles() function called during shutdown
3. Use atexit-style cleanup registration

**Additional context**
- Good first issue: straightforward resource cleanup
- Low severity (only affects debug mode, single FD leak)
- Pattern could apply to other debug loggers in codebase
- Comment at line 35 acknowledges "opened once via debugIMAPOnce to avoid leaking a descriptor per connection" but doesn't address lifetime leak

**Suggested fix:**
```go
// In fetcher package, add cleanup function:
func CloseDebugFiles() {
    if debugIMAPFile != nil {
        debugIMAPFile.Close()
    }
}

// In main package shutdown:
defer fetcher.CloseDebugFiles()
```

**OS**
All platforms

MD-Mushfiqur123:
## What?

Adds cleanup for the shared `DEBUG_IMAP` debug file handle.

The shutdown path now closes debug files before exiting, including code paths that call `os.Exit`, and logs any close error through `loglevel.Debugf`.

## Why?

Closes #715.

A deferred cleanup in `main()` is skipped when the process exits via `os.Exit`, which could leave the `DEBUG_IMAP` file descriptor open for the lifetime of the process. Closing it through the shared exit path makes cleanup reliable across CLI and daemon exits.


floatpanebot (dismissed):
Hi @MD-Mushfiqur123! Please fix the following issues with your PR:

- **Title**: Is too long (49 characters). The PR title must be strictly under 40 characters.

main.go
```diff
@@ -3947,6 +3947,7 @@ func main() {
 	args, level := parseGlobalFlags(os.Args)
 	os.Args = args
 	loglevel.Set(level)
+	defer fetcher.CloseDebugFiles()
```
FromSi: `defer` is not enough here because many paths in `main()` exit via `os.Exit(...)`, and Go does not run deferred calls on `os.Exit`. For example, `matcha daemon run` can open `DEBUG_IMAP`, then return through `runDaemonCLI` and hit `os.Exit(0)`, so `CloseDebugFiles()` is skipped.

FromSi (changes requested):

floatpanebot (dismissed):
Hi @FromSi! Please fix the following issues with your PR:

FromSi (approved):

andrinoff (approved):
lgtm

PR #1346, opened 2026-05-22
fix(fetcher): limit fetch chunks

### LINKED ISSUE #1106: BUG: Fetcher chunkSize over-fetches when caller asks for fewer than 50
### Describe the bug

`fetcher/fetcher.go:415-418`:

```go
chunkSize := limit
if chunkSize < 50 {
    chunkSize = 50
}
```

If the caller requests `limit=5`, we still fetch 50 from IMAP per loop, filter, sort, append, and trim to 5. That's a 10× waste of bandwidth and IMAP server load on every paginated tail.

### Expected behavior

Use `limit` directly; only inflate if the filter is known to drop more than half the messages (and bias the inflation to the historical hit rate).

puneetdixit200:
## What?
- remove the hard 50-message minimum chunk size in `FetchMailboxEmails`
- add a TLS IMAP recorder regression that verifies `limit=5` fetches only `96:100` from a 100-message mailbox

## Why?
When callers request a small page, the fetcher should not over-fetch 50 messages before filtering and trimming. This keeps IMAP bandwidth and server load proportional to the requested limit.

Fixes #1106

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -520,11 +520,7 @@ func FetchMailboxEmails(account *config.Account, mailbox string, limit, offset u
 
 	// Loop until we have enough emails or run out of messages
 	for len(allEmails) < int(limit) && cursor > 0 {
-		// Determine chunk size
 		chunkSize := limit
-		if chunkSize < 50 {
-			chunkSize = 50
-		}
 
 		from := uint32(1)
 		if cursor > chunkSize {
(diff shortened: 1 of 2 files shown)


floatpanebot (dismissed):
Hi @puneetdixit200! Please fix the following issues with your PR:

FromSi (approved):

andrinoff (approved):

PR #1347, opened 2026-05-22
fix: #1329 test confict

andrinoff:
## What?

Changes the test assertion due to #1329 changing the output to have `info: ` in front of it. Due to the PR not being rebased to have the #1316 commit.

## Why?

Tests fail.


floatpanebot (approved):

PR #1348, opened 2026-05-23
fix: remove signatures in general

andrinoff:
## What?

Removes the option to configure per-account settings in general

## Why?

This can already be achieved by going into the account settings. Doesn't need another menu for it


FromSi (approved):
lgtm

PR #1351, opened 2026-05-23
chore: sync gomod2nix.toml

floatpanebot:
## What?

Regenerates `gomod2nix.toml` to reflect the current `go.mod` / `go.sum`.

## Why?

Keeps the Nix build in sync with Go module changes. Without this, `nix build` fails when new or upgraded Go deps are missing from `gomod2nix.toml`. Generated automatically by the gomod2nix sync workflow.


andrinoff (approved):
lgtm

PR #1361, opened 2026-05-25
fix(view): sanitize HTML rendering

### LINKED ISSUE #653: BUG: HTML email rendering XSS vulnerability
Describe the bug: HTML emails rendered with incomplete sanitization. JavaScript URLs and event handlers may execute. Potentially dangerous with terminal emulators supporting OSC 8 hyperlinks.

To reproduce:
1. Receive HTML email with <a href="javascript:alert(1)">
2. Or onclick= event handler
3. If terminal supports OSC 8, could trigger actions
4. Risk varies by terminal emulator

Expected behavior: Strict HTML sanitization:
- Whitelist safe tags only (p, div, span, a, img, etc)
- Strip all javascript: URLs
- Remove all on* event attributes
- Sanitize CSS (no expression() or url() with javascript)
- Use bluemonday or similar library

Additional context:
- File: view/html.go, clib/htmlconv.c
- Currently converts to plaintext, but may miss cases
- Severity: Medium-High - XSS potential
- Complexity: High - security-critical HTML parsing

FromSi:
## What?

HTML email rendering now sanitizes untrusted content before converting it to terminal text.

Available after sanitization:
- safe text/layout tags such as paragraphs, lists, tables, headings, blockquotes, and inline formatting
- safe links using `http`, `https`, `mailto`, and `tel`
- image sources using `http`, `https`, safe `cid`, and valid `data:image` payloads

Not available after sanitization:
- `javascript:`, `vbscript:`, `file:`, relative, and protocol-relative links
- `data:` or `cid:` links in `<a href>`
- scripts, styles, event handlers like `onclick`, and unsafe attributes
- OSC 8 clickable image fallbacks for `cid:` and `data:image` sources

Demo of the fix:
image

## Why?

Closes #653

HTML emails could include unsafe links or attributes such as `javascript:` URLs and `on*` event handlers. In terminals with OSC 8 hyperlink support, these could become risky clickable terminal links.

This hardens HTML rendering for untrusted email content and addresses.

Demo of the issue:
image

### DIFF
--- a/internal/htmlsanitizer/lib_sanitizer.go
+++ b/internal/htmlsanitizer/lib_sanitizer.go
@@ -1,0 +1,54 @@
+package htmlsanitizer
+
+import (
+	"encoding/base64"
+	"net/url"
+	"regexp"
+
+	"github.com/microcosm-cc/bluemonday"
+)
+
+type LibSanitizer struct {
+	policy *bluemonday.Policy
+}
+
+func NewLibSanitizer() LibSanitizer {
+	return LibSanitizer{policy: newPolicy()}
+}
+
+func (s LibSanitizer) SanitizeBytes(html []byte) []byte {
+	return s.policy.SanitizeBytes(html)
+}
+
+func newPolicy() *bluemonday.Policy {
+	p := bluemonday.NewPolicy()
+	linkURLPattern := regexp.MustCompile(`(?i)^(https?://|mailto:|tel:)`)
+	imageURLPattern := regexp.MustCompile(`(?i)^(https?://|cid:|data:image/)`)
+	dataImagePrefixPattern := regexp.MustCompile(`(?i)^image/(gif|jpe?g|png|webp);base64,`)
+	p.AllowElements(
+		"a", "b", "blockquote", "br", "code", "div", "em", "h1", "h2",
+		"i", "img", "li", "ol", "p", "pre", "span", "strong", "table",
+		"tbody", "td", "th", "thead", "tr", "u", "ul",
+	)
+	p.AllowAttrs("href").Matching(linkURLPattern).OnElements("a")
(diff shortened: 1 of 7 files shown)


internal/htmlsanitizer/lib_sanitizer.go
```diff
@@ -0,0 +1,54 @@
...
+	p.AllowURLSchemeWithCustomPolicy("data", func(u *url.URL) bool {
+		if u.RawQuery != "" || u.Fragment != "" {
+			return false
+		}
+		prefix := dataImagePrefixPattern.FindString(u.Opaque)
+		if prefix == "" {
+			return false
+		}
+		_, err := base64.StdEncoding.DecodeString(u.Opaque[len(prefix):])
+		return err == nil
```
andrinoff: `StdEncoding` requires [RFC 4648](https://datatracker.ietf.org/doc/html/rfc4648) padding (=).

Better have `base64.RawStdEncoding` as a fallback, or decoding with `base64.NewDecoder(base64.StdEncoding, ...)` which is more lenient. The consequence of not fixing this is safe images silently disappearing in rendered emails.

view/html.go
```diff
@@ -755,6 +787,10 @@ func renderHTMLToText(htmlBody []byte, inline map[string]string, h1Style, h2Styl
 	return result, placements, nil
 }
 
+func isRemoteImageURL(src string) bool {
+	return strings.HasPrefix(src, "http://") || strings.HasPrefix(src, "https://")
+}
+
```
andrinoff, line 794: this is case sensitive, better use `toLower()` here too. `HTTPS://` is also a valid URL

andrinoff (changes requested):

andrinoff (approved):

PR #1362, opened 2026-05-25
feat: spellcheck and suggestions

### LINKED ISSUE #1160: FEAT: Spell-check in composer
### Describe the feature

Spell-check the composer body using `aspell` / `hunspell` if available, with squiggle highlighting and a key (`Ctrl+S`) to step through corrections.

### Why this matters

Real keyboard users send mail straight from the terminal. \"Send first, see typos in the sent folder\" is bad UX.

### Notes

- Shell out to `hunspell -l` (list misspelled) initially; we don't need an embedded engine.
- Skip checking inside markdown code blocks.

andrinoff:
## What?

Adds spellcheck library, `dict` CLI command, suggestions, and more

## Why?

Closes #1160

### DIFF
--- a/cli/dict.go
+++ b/cli/dict.go
@@ -1,0 +1,105 @@
+package cli
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/floatpane/matcha/spellcheck"
+)
+
+// RunDict dispatches `matcha dict <subcommand>`.
+func RunDict(args []string) error {
+	if len(args) == 0 {
+		return dictUsage()
+	}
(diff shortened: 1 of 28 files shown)


steveevansdev (approved):
valid

FromSi:
Right now `spellcheck.Highlight` processes the composer body line by line, and the composer calls it with `skipLine = -1`, so fenced code blocks are checked like normal prose. For example, words inside:

image

andrinoff:
yes, but code blocks are not usually written in emails, will add a detection for code, when we do the developer update (will come with v1). For now, it is consumer grade function, which can be turned off

PR #1367, opened 2026-05-27
fix(tui): settings focus nav

### LINKED ISSUE #1231: UX, Settings: several suggestions
- directly after opening the settings from the main menu makes it unclear what list has the focus, the left or the right - both look similar (green selection with **>**)
  - suggestion: the right side should not show a green selection if it does not have the focus
- support **->** to move focus from left to right menu
- moving from the left menu to the right allows to use the cursor-right key, but the it is not possible to get back to the left menu using the cursor-left key
  - support **<-** to move focus to left menu
  - support **Shift+Tab** to move focus to left menu
- align all the values in the right menu

FromSi:
## What?

Improves Settings pane focus navigation in the TUI.

- Adds `left` navigation back from the content pane to the settings menu.
- Allows `left` to leave Settings from the menu, like `esc`.
- Hides right-pane cursor/highlight when the content pane is not focused.
- Keeps bold styling only on the currently focused pane.
- Updates English Settings help text.
- Adds tests for horizontal pane focus and App Encryption input behavior.

view

## Why?

Closes #1231

Fixes confusing Settings focus behavior where both panes could appear selected at the same time.

This makes keyboard navigation clearer: `>` and bold styling now indicate the active focus, while `left` provides a consistent way to move back or exit.


i18n/locales/en.json
```diff
@@ -122,8 +122,8 @@
       "category_mailing_lists": "Mailing Lists",
       "category_encryption": "App Encryption",
       "category_plugins": "Plugins",
-      "help_menu": "↑/↓: navigate • right/enter: select • esc: go back",
-      "help_content": "esc: back to menu"
+      "help_menu": "↑/↓: navigate • right/enter: select • left/esc: go back",
+      "help_content": "left/esc: back to menu"
```
andrinoff, line 126: i think that escaping with left will lead to a lot of missclicks. let's leave just `esc`

CONTRIBUTING.md
```diff
@@ -60,6 +60,8 @@ type(scope): short description
 
 Common types: `feat`, `fix`, `docs`, `test`, `ci`, `chore`.
 
+Keep commit messages to 40 characters or fewer.
+
```
andrinoff, line 64: separate PR, please

i18n/locales/en.json
FromSi: Just to clarify: do you want me to remove `left` only as a shortcut for leaving the settings content pane, only for leaving the encryption screen, or should `left` no longer move back between panes at all?

andrinoff: left should not exit the settings menu, when you are in setting context selection

FromSi: fixed

andrinoff (dismissed):
great!

andrinoff (approved):

PR #1374, opened 2026-05-28
fix: prevent a fork bomb

andrinoff:
## What?

Add `view/main_test.go` with `TestMain` calling `termimage.MaybeRunWorker()` before `m.Run()`.

## Why?

termimage.DisplayWithSize` (called by `view.prerenderImage`) sandboxes image decode by spawning `os.Executable()` as subprocess with `TERMIMAGE_WORKER=1`. Worker subprocess MUST call `termimage.MaybeRunWorker()` first to intercept, decode, exit.

`main.go` does this for the real binary. Test binary did not. When `TestProcessBodyWithImageProtocol` ran Kitty/iTerm cases, `prerenderImage` spawned the test binary as worker. Child re-ran full test suite → triggered more `prerenderImage` calls → spawned more children → fork bomb. RAM exhausted, machine froze.

### DIFF
--- a/view/main_test.go
+++ b/view/main_test.go
@@ -1,0 +1,18 @@
+package view
+
+import (
+	"os"
+	"testing"
+
+	"github.com/floatpane/termimage"
+)
+
+// TestMain intercepts subprocess invocations made by termimage's sandbox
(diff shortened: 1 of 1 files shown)


floatpanebot (approved):

PR #1377, opened 2026-05-29
feat(composer): filter account picker

### LINKED ISSUE #554: FEAT: Search/Filter in Account Picker
Users with many email accounts might find it difficult to find the right account in the account picker list.

Add a simple search/filter functionality to the account picker in the composer (similar to how the inbox filtering works).

Scrolling through the list, which is slow for many accounts.

### Additional context
Implementation in `tui/composer.go`.

FromSi:
## What?

Added filtering to the composer account picker. Pressing the configured account picker filter key opens a filter input, narrows the account list, shows filtered result counts, and lets users apply or clear the filter before selecting an account.

view

## Why?

Closes #554

Users with many configured accounts need a faster way to find the correct sender account without scrolling through the full list.


andrinoff:
I know that this was a feature previously requested (transferred from non-OSS matcha). But, I don't think we need it yet at least, i'll keep it open

mavonx:
Hello @FromSi

I reviewed the code, and I have one mandatory change request and one optional suggestion (@andrinoff can decide whether to apply it or not).

- Mandatory: The email list should be circular.
- Optional: In the current implementation by @FromSi, when a user presses `f`, they are redirected to the filter page. After selecting a filter value and pressing `Enter`, the user is redirected back to the previous window with the filtered emails displayed.

My suggestion is slightly different: after the user presses `f` and selects a filter value, pressing `Enter` should keep the user on the filtered emails window instead of returning them to the previous one. At that point, the focus/cursor should automatically move to the filtered email list so the user can immediately navigate through the results.

demo

> Also, keep the account and filter counters as they are, but only on the filtering page (I removed them in the video by mistake).

andrinoff:
@mavonx @FromSi. In the current situation, where matcha is not optimized for more, than 4 accounts (in which case filtering is slower, than selecting), this is a bad feature. Let's not spend time on this right now. If we are to do this, I think we should first make matcha many-account-friendly (currently in the inbox, accounts overflow after 4, e.t.c.) 

Or maybe no user will ever need many accounts at once

andrinoff:
As well as that, the screen that you currently are showing, account picker, will just expand downwards and out of the screen on the amount of emails, where filtering is actually needed

andrinoff:
@FromSi move this PR to `release/v1` branch. We will include this in the v1, as we will have support for way more accounts.

Issue #1139, opened 2026-04-27
FEAT: Undo send (delayed delivery window)


andrinoff:
### Describe the feature

After a send action, hold the message in a local outbox for N seconds (configurable, default 10). The user can press `u` during that window to recall the message; otherwise the daemon delivers it.

### Why this matters

Standard Gmail/Outlook feature; saves users from \"reply-all to wrong list\" disasters.

### Notes

- Implementation overlaps heavily with FEAT scheduled send (#sibling). Same outbox, same daemon scanner, just a fixed delay.
- Provide a status-line countdown so the user knows the recall window is open.

floatpanebot:
This has had no activity for 30 days. It will be closed in 14 days unless updated. Comment or remove the `stale` label to keep it open.

PR #1378, opened 2026-05-30
fix: move email actions to daemon

### LINKED ISSUE #1255: FEAT: Background Archiving/Deleting/Moving

Currently, you have to sit on the "Moving email..." screen while the client sends the request and gets a success response

Move the emails via the daemon (introduced in #682), without showing the loading. Also, update the inbox view to display changes immediately, so that it doesn't feel like it took time. 

In case of an error, show it to user. 

Daemon is needed so that, in case the user closes Matcha, the sent request proceeds, instead of exiting prematurely.

_No response_

### Additional context

Related issue #1139, recommended to implement together.

mavonx:
## What?

Delegates email operations (delete, archive, move) to the daemon instead of blocking the UI with loading spinners. 

## Why?

Email operations can take time, especially for batch actions or slower IMAP connections. By delegating them to the daemon, the UI remains responsive, operations continue in the background, and users can safely close the TUI without interrupting the work.

Closes #1255


floatpanebot (dismissed):
Hi @mavonx! Please fix the following issues with your PR:

andrinoff:
@mavonx fixed

mavonx:
@andrinoff I pulled the updates and re-tested, but the delay is still there.

andrinoff:
@mavonx your previous daemon is likely running, `go run . daemon stop` and open the inbox

mavonx:
@andrinoff 

That's correct, I forgot to restart the daemon. Could you review the code before we merge?

main.go
```diff
@@ -2931,46 +2970,27 @@ func markEmailAsUnreadCmd(account *config.Account, uid uint32, accountID string,
 	}
 }
 
-func deleteFolderEmailCmd(account *config.Account, uid uint32, accountID string, folderName string, mailbox tui.MailboxKind) tea.Cmd {
+func (m *mainModel) deleteFolderEmailCmd(uid uint32, accountID string, folderName string, mailbox tui.MailboxKind) tea.Cmd {
 	return func() tea.Msg {
-		err := fetcher.DeleteFolderEmail(account, folderName, uid)
+		err := m.service.DeleteEmails(accountID, folderName, []uint32{uid})
```
andrinoff, line 2997: will panic if `m.service = nil`

main.go
```diff
@@ -1814,24 +1853,10 @@ func (m *mainModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) { //nolint:gocyclo
...
-		if m.folderInbox != nil {
-			m.folderInbox.RemoveEmail(msg.UID, msg.AccountID)
-			m.current = m.folderInbox
-			m.current, _ = m.current.Update(m.currentWindowSize())
-			return m, m.current.Init()
-		}
-		m.current = tui.NewChoice()
-		m.current, _ = m.current.Update(m.currentWindowSize())
-		return m, m.current.Init()
+		return m, nil
```
andrinoff, line 1858: The old code here did `m.current = m.folderInbox` + init. Now when a user reads an email and deletes it, `m.current` is never changed — they stay on the email reader viewing a deleted email. The email is removed from the list data, but the view doesn't navigate back to inbox.

main.go
```diff
@@ -1844,18 +1869,24 @@ func (m *mainModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) { //nolint:gocyclo
 		folderName := folderInbox
 		if m.folderInbox != nil {
 			folderName = m.folderInbox.GetCurrentFolder()
+			m.folderInbox.GetInbox().RemoveEmails(msg.UIDs, msg.AccountID)
```
andrinoff, line 1874: The batch handlers (lines 1872, 1902, 1932) only call `RemoveEmails` on the inbox widget and filter `m.folderEmails`, but never touch `m.emails`. Index-based lookups elsewhere will see stale results until the next full fetch.

andrinoff (changes requested):

andrinoff:
also i'd follow up with #1139, because, now 1 click means, that the message is deleted for sure.

mavonx:
@andrinoff Could you review it again now? I'll take care of #1139 in a separate PR.

andrinoff (approved):

PR #1382, opened 2026-05-30
chore: use udsrpc for daemon

andrinoff:
## What?

Replaced a lot of code with a new library [`go-uds-jsonrpc`](https://github.com/floatpane/go-uds-jsonrpc)

## Why?

A lot of code is reusable, and this may bring a lot of users into floatpane ecosystem. 

Also, makes it easier to maintain the parts, that are actually flawed


steveevansdev (approved):
lgtm

PR #1387, opened 2026-05-31
chore: use new library cardhl

andrinoff:
## What?

Replaces a lot of code with a new library [`go-openpgp-card-hl`](https://github.com/floatpane/go-openpgp-card-hl)

## Why?

Could be reused by other project and easier to maintain separately


floatpanebot (approved):

PR #1392, opened 2026-05-31
ci: remove release event

andrinoff:
## What?

Removes a release event trigger on nix-pkgs-bump workflow

## Why?

We need to rely on r-ryantm, because it will get merged sooner that way.


floatpanebot (approved):

PR #1395, opened 2026-06-01
chore: use floatpane/go-icalendar

andrinoff:
## What?

Replaces `calendar.go` with a new library [`go-icalendar`](https://github.com/floatpane/go-icalendar)

## Why?

Makes the code easier to expand and improve apart. Also, could be re-used by other people


floatpanebot (approved):

PR #1397, opened 2026-06-01
chore: use go-keybind

andrinoff:
## What?

Uses [`go-keybind`](https://github.com/floatpane/go-keybind).

## Why?

Easier to maintain/expand


floatpanebot (approved):

PR #1398, opened 2026-06-01
fix(badge): dedupe unread count

### LINKED ISSUE #1107: BUG: syncUnreadBadge double-counts mail in both stores
### Describe the bug

`main.go:179-200` walks both `m.emailsByAcct` and `m.folderEmails` and increments `count` for every unread email seen. The two stores overlap — `emailsByAcct` is populated from the same fetched messages that go into `folderEmails`. The dock badge therefore shows roughly 2× the real unread count.

### Expected behavior

Count from a single source of truth. If both stores are needed for other reasons, deduplicate by `{AccountID, UID}` before counting.

FromSi:
## What?

Deduplicate unread badge counting across `emailsByAcct` and `folderEmails` by tracking seen emails with `AccountID + UID`.

Added a regression test for the case where the same unread email exists in both stores.

image

## Why?

Closes #1107

`syncUnreadBadge` counted unread emails from both stores independently, but the stores can contain the same fetched messages. This could make the macOS unread badge show roughly double the real unread count.

image

### DIFF
--- a/main.go
+++ b/main.go
@@ -242,26 +242,45 @@ func waitForLogEntry(ch <-chan logging.Entry) tea.Cmd {
 	}
 }
 
-func (m *mainModel) syncUnreadBadge() {
-	if runtime.GOOS != goosDarwin {
-		return
-	}
+func unreadBadgeCount(emailsByAcct, folderEmails map[string][]fetcher.Email) int {
 	count := 0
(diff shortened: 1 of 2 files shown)


andrinoff (approved):
lgtm

Issue #1411, opened 2026-06-02
git-mail integration


andrinoff:
Send patches via email (`git send-email`-style) and apply incoming patches (`git am`) directly from matcha. Detect patch emails, offer apply-to-repo action.

floatpanebot:
This has had no activity for 30 days. It will be closed in 14 days unless updated. Comment or remove the `stale` label to keep it open.

PR #1420, opened 2026-06-03
feat(gitmail): initial

andrinoff:
## What?

Adds a git-mail apply workflow to matcha: take a `git format-patch` / `git send-email` message and apply it to a local working tree, without shelling out to git.

## Why?

Part of #1411 (not fully done)


floatpanebot (approved):

PR #1427, opened 2026-06-03
fix: gate status request

### LINKED ISSUE #1426: Folder list fails on Proton Bridge and Exchange Online (LIST-STATUS requested without checking the capability)
I have three accounts set up: Gmail, Proton (via Proton Mail Bridge, configured as a custom IMAP server), and an Exchange Online / Office 365 account. Inbox and sending work on all of them, but folder listing fails on the two non-Gmail accounts:

```
Folder fetch failed for 2 account(s):
  <exchange>: imap: BAD Command Argument Error. 12
  <proton>: in response: cannot read tag: imapwire: expected '*' or atom, got ' '
```

I turned on DEBUG_IMAP to see what was actually going over the wire, and the common factor is the LIST-STATUS extension. FetchFolders (in fetcher/fetcher.go) always asks for the unseen count inline:

```go
listCmd := c.List("", "*", &imap.ListOptions{
    ReturnStatus: &imap.StatusOptions{
        NumUnseen: true,
    },
})
```

go-imap turns that into `LIST "" "*" RETURN (STATUS (UNSEEN))`, which requires LIST-STATUS (RFC 5819). Nothing checks that the server advertises it first, and the two accounts that break are the two that don't:

Proton Mail Bridge 3.24.2:
```
* CAPABILITY AUTH=PLAIN ID IDLE IMAP4rev1 MOVE STARTTLS UIDPLUS UNSELECT
```
Exchange Online:
```
* CAPABILITY IMAP4 IMAP4rev1 AUTH=PLAIN AUTH=XOAUTH2 SASL-IR UIDPLUS MOVE ID UNSELECT CHILDREN IDLE NAMESPACE LITERAL+
```
Gmail (works):
```
* CAPABILITY IMAP4rev1 ... LIST-EXTENDED LIST-STATUS ... SPECIAL-USE ...
```

Exchange rejects the RETURN argument outright (BAD), and Proton Bridge sends back a reply go-imap can't parse. A plain `LIST "" "*"` works fine on both, it's only the inline STATUS that trips them up.

Gating it on the capability should be enough:

```go
opts := &imap.ListOptions{}
if c.Caps().Has(imap.CapListStatus) {
    opts.ReturnStatus = &imap.StatusOptions{NumUnseen: true}
}
listCmd := c.List("", "*", opts)
```

and if the unseen counts matter on those servers, fall back to a STATUS per mailbox.

To reproduce: add an account on any server that doesn't advertise LIST-STATUS (Proton Bridge, or Exchange / Office 365) and open the folder list.

Version: matcha 0.40.1, go-imap v2, Linux.

andrinoff:
## What?

Gate the `LIST ... RETURN (STATUS (UNSEEN))` request in `FetchFolders` on the server advertising the LIST-STATUS capability (RFC 5819). When the server doesn't support it, send a plain `LIST "" "*"` and populate unread counts with a per-mailbox `STATUS (UNSEEN)` fallback instead.

## Why?

Fixes #1426

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -1919,11 +1919,18 @@ func FetchFolders(account *config.Account) ([]Folder, error) {
 	}
 	defer c.Close() //nolint:errcheck
 
-	listCmd := c.List("", "*", &imap.ListOptions{
-		ReturnStatus: &imap.StatusOptions{
-			NumUnseen: true,
-		},
-	})
+	// Only request the unseen count inline via LIST ... RETURN (STATUS (UNSEEN))
+	// when the server advertises LIST-STATUS (RFC 5819). Servers without it
+	// (e.g. Proton Mail Bridge, Exchange Online) either reject the RETURN
+	// argument outright or send a reply go-imap can't parse (#1408). For those
+	// we fall back to a per-mailbox STATUS below.
+	hasListStatus := c.Caps().Has(imap.CapListStatus)
+	listOpts := &imap.ListOptions{}
+	if hasListStatus {
+		listOpts.ReturnStatus = &imap.StatusOptions{NumUnseen: true}
+	}
+
+	listCmd := c.List("", "*", listOpts)
 	defer listCmd.Close() //nolint:errcheck
 
 	var folders []Folder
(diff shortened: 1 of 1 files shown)


subtervisor:
Can confirm, was seeing this with my email provider as well. The PR binaries successfully fix the issue for me.

floatpanebot (approved):

PR #1434, opened 2026-06-04
feat: send or unsend matcha messages

### LINKED ISSUE #1139: FEAT: Undo send (delayed delivery window)

mavonx:
## What?

Messages are held for `N` seconds before delivery, with a recall window.

## Why?

Saves users from sending mistakes, just like Gmail/Outlook.

Closes #1139

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -129,6 +129,7 @@ type Config struct {
 	DateFormat              string        `json:"date_format,omitempty"`
 	Language                string        `json:"language,omitempty"` // Language code (e.g., "en", "es", "de")
 	BodyCacheThresholdMB    int           `json:"body_cache_threshold_mb,omitempty"`
+	UndoDelaySeconds        int           `json:"undo_delay_seconds,omitempty"`
 	// PluginSettings stores user-configurable values for installed plugins,
 	// keyed by plugin name then setting key. Values are JSON-native types
 	// (bool, float64, string) matching the plugin's declared schema.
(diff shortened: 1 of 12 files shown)


andrinoff (changes requested):
I'll review later changes in the evening

mavonx:
> 1. Inline images are broken, when sent to QueueEmail (not passed)
> 2. I'm pretty sure, that the daemon has no countdown, it just fires delete, even though UI says that it's going to do it in 10 seconds
> 3. Keybind should be in `keybinds.go` (reconfigurable)
> 4. Maybe, configure delays in the config
> 5. A nitpick, but, if the daemon crashes in the state of "Sending in n seconds" queued emails will not be sent

@andrinoff

Regarding the second point, I see two possible approaches:

1. Remove the countdown from the TUI and keep it only in the daemon. In this case, we won't have a countdown displayed in the TUI; I'll simply show `"Sending... (u to undo)"`. I recommend this approach, as it's similar to how Gmail handles it in their [demo](https://support.google.com/mail/answer/2819488?hl=en&co=GENIE.Platform%3DDesktop).
2. Remove the countdown logic from the TUI and have the TUI request the countdown value from the daemon in order to display it.

andrinoff:
Whichever you want, maybe a timer looks like a ticking bomb and Gmail/iCloud mail style would be better. also, I'd like to add to the configurable countdown, that it should be around 5 seconds default, not 10

andrinoff (changes requested):
@mavonx `daemonrpc.SendEmailParams` still has no `Images` field, so `sendOutboxEntry` calls `sender.SendEmail(..., nil, attachments, ...)`

andrinoff (approved):
LGTM

andrinoff:
@mavonx can you backport manually? instructions are above, open a PR afterwards. If you have questions on how to do it, please, ask in the dev channel, not on the issue

PR #1435, opened 2026-06-04
chore: use bubble-overlay v0.1.0

andrinoff:
## What?

Uses `bubble-overlay v0.1.0` to simplify the code of the command palette. 

## Why?

This will be reusable in other floating modals.


floatpanebot (approved):

PR #1436, opened 2026-06-04
fix: add row offset for split pane mode

andrinoff:
## What?

Adds a row offset for displaying images in a vertical split pane mode

## Why?

Before:

Screenshot 2026-06-04 at 12 45 50

After:

Screenshot 2026-06-04 at 12 45 12


steveevansdev (approved):
lgtm

PR #1442, opened 2026-06-05
fix: move "Language set" to loglevel

LeaWhoCodes:
## What?

Moves the log message for i18n to loglevel `verbose`

## Why?

Each time you used to load matcha in secure mode, these messages would appear and possibly block the UI


floatpanebot (dismissed):
Hi @LeaWhoCodes! Please fix the following issues with your PR:

andrinoff (approved):
lgtm

PR #1445, opened 2026-06-05
fix: JMAP integration

masukomi:
## What?
The code to support JMAP authentication via API tokens was in place, it just wasn't wired up correctly. This fixes that.

- JMAP protocol selection ended up getting routed to IMAP code
- no way to specify `token` for `account.AuthMethod`
- `hasBackendProvider` didn't account for/support `jmap`
- modified placeholder when `account.AuthMethod` is `token` to say `API Token` instead of `Password / App Password`
- modified tip for `inputAuthMethod` when protocol is `jmap`

## Why?
Because I use Fastmail, want to use JMAP, and want to be able to read my email.

JMAP authentication with Fastmail via API Tokens wasn't working for multiple reasons. Fastmail is the largest (only?) JMAP supporting email host and they require API Tokens. 

I have manually confirmed that this authenticates just fine with my Fastmail account via JMAP and the API token.

### DIFF
--- a/backend/jmap/jmap.go
+++ b/backend/jmap/jmap.go
@@ -53,7 +53,7 @@ func New(account *config.Account) (*Provider, error) {
 		SessionEndpoint: account.JMAPEndpoint,
 	}
 
-	if account.AuthMethod == "oauth2" {
+	if account.AuthMethod == "oauth2" || account.AuthMethod == "token" || account.AuthMethod == "" {
 		client.WithAccessToken(account.Password)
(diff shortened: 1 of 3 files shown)


floatpanebot (dismissed):
Hi @masukomi! Please fix the following issues with your PR:

floatpanebot (dismissed):
Hi @masukomi! Please fix the following issues with your PR:

andrinoff (approved):
lgtm!

PR #1446, opened 2026-06-05
fix: JMAP fetching and moving

masukomi:
## What?
in `jmap.go`

- added a `resolveUID` function
- added a `resolveUIDByQuery` function

## Why?
Email body fetch always failed with "jmap: no cached ID" (backend/jmap/jmap.go)
FetchEmailBody relied on p.idToJMAPID being pre-populated by FetchEmails on the same Provider instance. However, the fetcher layer creates a fresh Provider for every call, so the map is always empty when FetchEmailBody runs.

The fix adds a resolveUID method that checks the in-memory cache first (fast path when the daemon reuses the same instance), then falls back to resolveUIDByQuery. The fallback issues an Email/query for the folder, reads the JMAP string IDs directly from QueryResponse.IDs, hashes each one with FNV-32a, and returns the match — also warming the cache as a side effect.

### DIFF
--- a/backend/jmap/jmap.go
+++ b/backend/jmap/jmap.go
@@ -362,8 +362,8 @@ func (p *Provider) FetchAttachment(_ context.Context, _ string, _ uint32, partID
 	return io.ReadAll(reader)
 }
 
-func (p *Provider) MarkAsRead(_ context.Context, _ string, uid uint32) error {
-	jmapID, err := p.lookupJMAPID(uid)
+func (p *Provider) MarkAsRead(_ context.Context, folder string, uid uint32) error {
+	jmapID, err := p.resolveUID(folder, uid)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


floatpanebot (dismissed):
Hi @masukomi! Please fix the following issues with your PR:

- **Title**: Is too long (48 characters). The PR title must be strictly under 40 characters.

floatpanebot (dismissed):
Hi @masukomi! Please fix the following issues with your PR:

- **Title**: Is too long (41 characters). The PR title must be strictly under 40 characters.

masukomi:
added the commit about moving things. Will check in on the CI results later.

andrinoff:
there are some issues that will need further follow ups. but for now, this is lgtm.


To not forget, follow ups:

- [ ] Configurable limit of UID (currently set at hardcoded 10k)
- [ ] Potentially expensive fallback runs on the hot path for every per-UID action. (`resolveUIDByQuery` performs a full `Email/query` over the entire folder (up to 10k IDs) and hashes every single one whenever the cache misses.)

andrinoff (approved):

PR #1447, opened 2026-06-05
feat: setup guide

andrinoff:
## What?

Adds a setup guide prompting to install additional features, shows keybinds and key features

## Why?

It may be confusing at first, and this may catch attention of new users


steveevansdev (approved):
lgtm

Issue #1448, opened 2026-06-06
BUG: Screen not fully redrawn in some views


Brixy:
### Describe the bug

The TUI is not fully redrawn (not quite sure about the correct wording) when entering "View Inbox", please see screenshots below.

Other views (compose, marketplace, settings) are not affected.

I'm using matcha's encryption, so this may be the reason. Currently I cannot turn encryption off due to #684 for testing purposes.

**Two ways to force redrawing the TUI**

1. When I go back to matcha's start screen (`ESC`) and re-enter inbox view everything looks correct.

2. I'm using Niri or River as windows managers. As soon as I resize the window containing matcha, matcha's TUI is redrawn and everything looks correct.

### To reproduce

1. Enable matcha's encryption(?)
2. Kill matcha
3. Start matcha
4. Enter password
5. Enter "View Inbox" -> screen looks like shown in the screenshots
6. `ESC`
7. Enter "View Inbox" -> everything looks correct

### Expected behavior

I would expect the TUI screen to be fully redrawn when entering any of matcha's views.

Remaining text from the start screen should be removed.

Text like ` / __ '__ \/ __ '/ __│ ___/ __ \/ __ '`  should be removed.

### Screenshots

<img width="420" height="478" alt="Image" src="https://github.com/user-attachments/assets/9aa4282b-9ccb-44b5-860a-8f131c744bb2" />
<img width="418" height="151" alt="Image" src="https://github.com/user-attachments/assets/564f90f1-83bd-4bf5-9e45-00cd67338c54" />
<img width="1505" height="142" alt="Image" src="https://github.com/user-attachments/assets/78c34757-3fc8-4488-81bb-bb9382d04fd3" />

### Matcha version

496a775737038e99642775b3f32571dcc6f48127

### Go version

go1.26.4 linux/amd64

### OS

Linux

### OS version

Void

### Architecture

amd64 (x86_64)

### Terminal

foot

andrinoff:
@Brixy is this the first version where this occurs? Seems like a weird bug, all elements are stacking on top of each other for the whole screen, leaving almost no space. 

I did see this bug on some older version exclusively on Windows, but this is definitely not the case

Brixy:
This issue has occurred in every version I've used so far. Just wanted to report when I realized that you are working von Matcha 1.0. (Looking forward!)

Would it help if I tested Matcha on standard DE, just in case the issue is caused by tiling or floating window managers? I could test it on something like XFCE on short notice.

[...]

PR #1450, opened 2026-06-07
feat: make daemon optional

andrinoff:
## What?

Makes Background daemon optional in the settings

## Why?

Temporary fix for #1448, also a good option to have for users to turn off, if they do not want the daemon on.


steveevansdev (approved):
lgtm

PR #1455, opened 2026-06-07
fix: make undo message non-blocking

andrinoff:
## What?

Improves queue, by making undo non-blocking

## Why?

You had to wait for 5 seconds in order to use Matcha again


floatpanebot (approved):

PR #1482, opened 2026-06-17
chore: add make install

andrinoff:
## What?

Add `make install` that overrides the current matcha install with the newly built

## Why?

Adds an option for people to clone the repo and install a freshly built, instead of downloading binaries, and developers to use the latest version for day-to-day testing.


floatpanebot (approved):

PR #1484, opened 2026-06-17
ci: panic errors in rust and flatpak

andrinoff:
## What?

Adds Rust requirement in flatpak, and mute panic errors (null guarded code) 

## Why?

Releases fail otherwise


floatpanebot (approved):

PR #1485, opened 2026-06-17
fix: adapt makefile for rust

andrinoff:
## What?

Adapts makefile to run with Rust

## Why?

All previous commands were failing, because `built-rust` was not passed


floatpanebot (approved):

PR #1491, opened 2026-06-19
fix(tui): sync unread badge on delete

youdie006:
## What?

After deleting or archiving emails, the delete/archive `Update` handlers (`DeleteEmailMsg`, `ArchiveEmailMsg`, `BatchDeleteEmailsMsg`, `BatchArchiveEmailsMsg`) now call `m.syncUnreadBadge()`, mirroring the read/unread handlers, so the unread badge recomputes immediately. `syncUnreadBadge` also caches the count on every OS so the behavior is unit-testable. Adds four tests asserting the macOS badge drops to 0 after the last unread email is deleted/archived.

## Why?

The delete/archive handlers mutated the unread stores but never recomputed the unread menu bar badge, so the counter only refreshed on the next fetch.

### DIFF
--- a/main.go
+++ b/main.go
@@ -145,6 +145,9 @@ type mainModel struct {
 	sendNotice    string
 	pendingAction *pendingEmailAction
 	actionNotice  string
+	// unreadBadge caches the unread count last pushed to the OS badge so the
+	// value the badge derives from is observable after email operations.
+	unreadBadge int
 }
 
 type logEntryMsg struct {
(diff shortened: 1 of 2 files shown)


floatpanebot (dismissed):
Hi @youdie006! Please fix the following issues with your PR:

andrinoff:
@youdie006 thank you for your contribution. 

The issue was not about this badge, but rather the folder unread. Unlinking the issue, I'll check out the PR a bit later.

PR #1615, opened 2026-06-23
chore: delete patch file

andrinoff:
## What?

Removes patch file, accidentally added

## Why?
Sent via git-mail


floatpanebot (approved):

PR #1616, opened 2026-06-23
chore: change auto-update for RCs

andrinoff:
## What?

Changes the release workflow (rc) to publish to brew, and auto-updater to detect updates correctly, and update via homebrew, or snapcraft. 

## Why?
So that auto-updater does not try to find the versions elsewhere


floatpanebot (approved):

PR #1621, opened 2026-06-24
fix: add cap to image rendering

andrinoff:
## What?

Adds a cap to image rendering, so that the images dont break the whole TUI

## Why?

Previously, big images would make the whole screen black, when scrolled.


floatpanebot (approved):

PR #1622, opened 2026-06-24
feat: install themes via CLI

andrinoff:
## What?

Adds theme installation to `matcha install` CLI command to support the new matcha-themes store of themes.

## Why?

Previously, you had to download the themes on your computer and move them to the themes folder manually.


steveevansdev (approved):
lgtm

PR #1628, opened 2026-06-25
chore: removes onboarding from gif demo

andrinoff:
## What?

Adds `has_seen_setup_guide` parameter to the config file of demo GIF workflow.

## Why?

Currently generated GIF contains the set up guide, and it is unnecessary in the demo.


floatpanebot (approved):

PR #1633, opened 2026-06-28
feat: ui customization via plugins

andrinoff:
## What?

Adds UI customization via the Plugin API

## Why?

Some users may want to override the text, banner, e.t.c.


steveevansdev (approved):
lgtm

PR #1636, opened 2026-06-29
fix: attachments being dropped

andrinoff:
## What?

Fixes dropped attachments

## Why?

Attachments were being dropped, caused by the refactoring in #1603


floatpanebot (approved):

PR #1637, opened 2026-06-30
fix: better log for plugins

andrinoff:
## What?

Passes successful logs (plugin loaded) through verbose `loglevel` (activate with `matcha --verbose` and passes errors to notifications instead of `stderr`

## Why?

The successes are not needed to see all the time and the terminal becomes cluttered, when there are a lot of them.

Errors are more needed inside the terminal, rather, than in `stderr`

### DIFF
--- a/plugin/load_notify_test.go
+++ b/plugin/load_notify_test.go
@@ -1,0 +1,124 @@
+package plugin
+
+import (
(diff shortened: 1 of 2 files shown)


steveevansdev (approved):
lgtm

PR #1639, opened 2026-06-30
fix: better view UI

andrinoff:
## What?

Improves UI in the email view

## Why?

Previous UI was confusing at times, with too much clutter on 1 line


andrinoff:
.

PR #1643, opened 2026-07-02
fix: resolve sent folders

### LINKED ISSUE #1641: Saving to Sent fails on non-English Exchange/Office 365 (APPEND uses a hardcoded folder name instead of the \Sent special-use attribute)
Same three accounts as in #1426: Gmail, Proton (via Proton Mail Bridge, custom IMAP), and an Exchange Online / Office 365 account. Sending works everywhere, but on the Office 365 account matcha can't save the sent copy. Accepting or declining a calendar invite goes out fine, then this shows up:

```
2026/07/02 10:21:03 Failed to append RSVP to Sent folder: imap: NO [TRYCREATE] Folder is not found.
                                                             RSVP sent: ACCEPTED
```

The reply itself is delivered (RSVP sent: ACCEPTED), it's only the APPEND to Sent that fails.

The name matcha appends to comes from a hardcoded table in fetcher/fetcher.go:

```go
func getSentMailbox(account *config.Account) string {
	switch account.ServiceProvider {
	case config.ProviderGmail:
		return "[Gmail]/Sent Mail"
	case "outlook":
		return "Sent Items"
	case "icloud":
		return "Sent Messages"
	default:
		return "Sent"
	}
}
```

and AppendToSentMailbox uses that string directly, without asking the server what the Sent folder is actually called:

```go
sentMailbox := getSentMailbox(account)
appendCmd := c.Append(sentMailbox, ...)
```

The catch is that Exchange localizes the special folder names to the mailbox language. My Office 365 mailbox is Dutch, so the Sent folder isn't "Sent Items", it's "Verzonden items". "Sent Items" doesn't exist, so the APPEND comes back NO [TRYCREATE]. The same thing would hit French ("Éléments envoyés"), German ("Gesendete Elemente"), and so on, plus any custom IMAP account whose Sent folder isn't literally named "Sent".

Same shape as #1426, really: matcha assumes a name instead of asking the server for it. And the fix is already sitting in the same file. getMailboxByAttr looks a mailbox up by its RFC 6154 special-use attribute, and the Trash path already uses it, falling back to the hardcoded name only if the lookup fails:

```go
trashMailbox, err := getMailboxByAttr(c, imap.MailboxAttrTrash)
if err != nil {
	trashMailbox = getTrashMailbox(account)
}
```

AppendToSentMailbox just never got the same treatment. Doing the same with \Sent finds the real folder whatever the language:

```go
sentMailbox, err := getMailboxByAttr(c, imap.MailboxAttrSent)
if err != nil {
	sentMailbox = getSentMailbox(account)
}
```

That covers the RSVP path and the normal send path in one place (both go through AppendToSentMailbox), and ArchiveSentEmail has the same blind spot since it also calls getSentMailbox. The send succeeds either way, so it's easy to miss, the copy just quietly never shows up in Sent.

To reproduce: add an Exchange / Office 365 account whose mailbox language isn't English (or any custom IMAP account whose Sent folder isn't named "Sent") and send something, or RSVP to an invite.

I can test a fix across all three accounts like I did for #1426.

Version: matcha 0.43.0, go-imap v2, Linux.

andrinoff:
## What?

All Sent mailbox lookups now resolve via `getMailboxByAttr(c, imap.MailboxAttrSent)` with fallback to the hardcoded `getSentMailbox`, matching the existing Trash/Archive pattern.

## Why?

Fixes #1641

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -538,7 +538,13 @@ func FetchMailboxEmails(account *config.Account, mailbox string, limit, offset u
 	if fetchEmail == "" {
 		fetchEmail = strings.ToLower(strings.TrimSpace(account.Email))
 	}
-	isSentMailbox := mailbox == getSentMailbox(account)
+	// Resolve the sent mailbox name dynamically so localized names (e.g.
+	// "Verzonden items" on Dutch Exchange) are recognized as the Sent folder.
+	sentMailbox, err := getMailboxByAttr(c, imap.MailboxAttrSent)
+	if err != nil {
+		sentMailbox = getSentMailbox(account)
+	}
+	isSentMailbox := mailbox == sentMailbox
 
 	// Delivery header section for matching auto-forwarded emails
 	deliveryHeaderSection := &imap.FetchItemBodySection{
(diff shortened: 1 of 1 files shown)


floatpanebot (approved):

PR #1648, opened 2026-07-04
fix: daemon socket death

andrinoff:
## What?

Fixes 2 bugs causing connection close and daemon socket death.

1. Added a `call()`  wrapper around all `daemonService` methods that:

* Detects connection-level errors via `isConnError()`
* Closes the dead client and calls `tryConnectWithConfig()` to reconnect (auto-starts
daemon if needed)
* Retries the operation once on the fresh connection

2. Added `closeMu sync.Mutex` to serialize the `close(done)` operation between `Close()` and `readLoop()`.

## Why?

When socket died, matcha became unusable for 5-10 minutes


floatpanebot (approved):

PR #1649, opened 2026-07-04
fix: release fail

andrinoff:
## What?

Replaces `ar rcs` --> `zig ar rcs`

## Why?

Workflows failed previously


floatpanebot (approved):

PR #1657, opened 2026-07-05
fix: threading issues

andrinoff:
## What?

Improves threaded view, making the first email (latest) accessible

## Why?

Previously, you could not access the last email


floatpanebot (approved):

PR #1668, opened 2026-07-09
fix(daemonclient): register IMAP/POP3

2005czq:
## What?

Register IMAP and POP3 backends in daemonclient's directService path.

## Why?

daemonclient already imports JMAP and Maildir for directService backend registration. IMAP and POP3 are supported by the app as well, but are not registered locally in daemonclient.

This keeps the direct fallback backend set consistent with the supported protocols.


andrinoff (approved):
LGTM.

PR #1680, opened 2026-07-13
feat(cli): add cli subcommands

### LINKED ISSUE #1642: FEAT: Round out the non-interactive CLI (read + manage mail, not just send)

The CLI docs open with "these work without launching the TUI and are ideal for scripts, cron jobs, and AI agent integration," but on the mail side the only thing you can actually do non-interactively today is `matcha send`. Reading and managing mail still needs the TUI, so there's no way to script "list what's in this folder", "print this message", or "archive/mark these" without driving the interactive UI. That's the gap between what the docs promise and what's exposed.

Expose the read and manage operations as subcommands, following the same pattern `matcha send` already uses. I'm filing this as one request rather than piecemeal because the groundwork is basically all there: the operations already exist as plain package-level functions in `fetcher/` that take a `*config.Account` and run directly (no daemon needed), which is what `runSendCLI` calls today. For example:

- `fetcher.FetchFolders`
- `fetcher.FetchFolderEmails` / `FetchFolderEmailBody`
- `fetcher.MarkEmailAsReadInMailbox` / `MarkEmailAsUnreadInMailbox`
- `fetcher.ArchiveEmailFromMailbox` / `DeleteEmailFromMailbox` / `MoveEmailsToFolder`
- `fetcher.SearchMailbox`

So a first cut is mostly flag parsing plus an output formatter. Rough shape, mirroring `matcha send`:

```
matcha folders [--account <email>]
matcha list    --folder INBOX [--account <email>] [--limit 50] [--offset 0]
matcha read    --folder INBOX --uid <n> [--account <email>]
matcha search  --folder INBOX --query "..." [--account <email>]
matcha mark-read | mark-unread --folder INBOX --uid <n> [--account <email>]
matcha archive | delete --folder INBOX --uid <n> [--account <email>]
matcha move    --uid <n> --from INBOX --to Archive [--account <email>]
```

A `--json` option on the read commands (`list`, `read`, `folders`, `search`) would make this genuinely usable from scripts and agents, which the docs already name as a goal. Default to a human-readable table, JSON when asked. Account selection could reuse the same `--from` matching `send` already has (login + fetch email).

Scripting the TUI with expect-style input (brittle, breaks on any UI change), or talking to the daemon socket directly (`daemonrpc` is internal and not a stable contract). Neither is a real substitute for first-class subcommands.

### Additional context

`runSendCLI` in `main.go` is the template to copy (parse flags, `LoadConfig`, resolve account, call the fetcher/sender function, print), and new commands slot into `cli/` next to `config.go`, `contacts_export.go`, `dict.go`, etc., wired through the `os.Args` dispatch in `main()`. matcha also has the `daemonclient.Service` interface (used by the TUI and daemon) wrapping these same operations mode-agnostically, if routing the CLI through that is preferred for daemon-awareness. Either way the underlying operations already exist, so this is mostly surface work.

Before I build anything though I'd like your read on it: is this something you'd want, and is the command/flag shape roughly right? If it looks good to you I'll take it on and open a PR.

Version: matcha 0.43.0.

Robin-Everaars:
## What?

Adds a non-interactive CLI to Matcha so mail can be handled from scripts and one-off commands without entering the TUI. New subcommands:

- `list <folder>`: list messages in a folder (tabular, or `--json`)
- `read <uid>`: print a message body; `--json` includes attachment metadata
- `search <query>`: search using the same query DSL the TUI uses
- `folders`: list available folders
- mail management: `archive`, `delete`, `mark-read`, `mark-unread`, `move`

Supporting changes:

- **daemon**: new `Search` RPC (service method + handler). `NewCLIClient(cfg, autoStart)` lets a one-shot invocation use a running daemon when present and otherwise fall back to a direct connection, rather than forking a daemon.
- **backend**: `Attachment` gains explicit snake_case JSON tags and excludes the raw `Data` bytes, so scripted `--json` consumers never receive megabytes of base64.
- **fix**: the `mark-read` handler previously logged per-UID failures but reported success over RPC; it now joins and returns them, matching the other mutating handlers and direct-mode semantics.
- **refactor**: commands write through hookable `Out`/`ErrOut` writers; the existing `contacts-export`, `config`, and `upgrade` commands were moved onto the same writers for consistency and testability. Shared plumbing (account resolution by `--from`, tabular/JSON rendering, folder-name normalization, interspersed flag parsing) lives in `cli/util.go`.

Every new command ships with unit tests.

## Why?

Matcha is TUI-only today, which makes it awkward to use from scripts, cron jobs, or other tooling. The CLI docs already promise these commands "work without launching the TUI and are ideal for scripts, cron jobs, and AI agent integration," but on the mail side only `matcha send` was actually exposed. This closes that gap: the core mailbox operations (list / read / search / folders, plus manage) become non-interactive subcommands, making Matcha scriptable and composable with the rest of a shell workflow, while reusing the existing daemon and query DSL so behaviour stays consistent with the TUI.

Closes #1642.

### DIFF
--- a/backend/backend.go
+++ b/backend/backend.go
@@ -91,19 +91,19 @@ type Email struct {
 
 // Attachment holds data for an email attachment.
 type Attachment struct {
-	Filename         string
-	PartID           string
-	Data             []byte
-	Encoding         string
-	MIMEType         string
-	ContentID        string
-	Inline           bool
-	IsSMIMESignature bool
-	SMIMEVerified    bool
-	IsSMIMEEncrypted bool
-	IsPGPSignature   bool
-	PGPVerified      bool
-	IsPGPEncrypted   bool
+	Filename         string `json:"filename"`
+	PartID           string `json:"part_id"`
+	Data             []byte `json:"-"` // Don't include raw data in JSON by default
+	Encoding         string `json:"encoding"`
+	MIMEType         string `json:"mime_type"`
+	ContentID        string `json:"content_id"`
+	Inline           bool   `json:"inline"`
+	IsSMIMESignature bool   `json:"is_smime_signature"`
+	SMIMEVerified    bool   `json:"smime_verified"`
+	IsSMIMEEncrypted bool   `json:"is_smime_encrypted"`
+	IsPGPSignature   bool   `json:"is_pgp_signature"`
+	PGPVerified      bool   `json:"pgp_verified"`
+	IsPGPEncrypted   bool   `json:"is_pgp_encrypted"`
 }
 
 // SearchQuery is the parsed form of a user query string.
--- a/cli/config.go
+++ b/cli/config.go
@@ -38,6 +38,6 @@ func RunConfig(args []string) error {
 	cmd := exec.Command(editor, target) //nolint:gosec,noctx
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
+	cmd.Stderr = ErrOut
 	return cmd.Run()
 }
--- a/cli/contacts_export.go
+++ b/cli/contacts_export.go
@@ -21,19 +21,19 @@ func RunContactsExport(args []string) error {
 	noHeader := fs.Bool("no-header", false, "omit CSV header row")
 
 	fs.Usage = func() {
-		fmt.Fprintln(Out, "Usage: matcha contacts export [flags]")
-		fmt.Fprintln(Out, "")
-		fmt.Fprintln(Out, "Export contacts from cache to JSON or CSV format.")
-		fmt.Fprintln(Out, "")
-		fmt.Fprintln(Out, "Flags:")
+		fprintln(Out, "Usage: matcha contacts export [flags]")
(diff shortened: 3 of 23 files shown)


floatpanebot (dismissed):
Hi @Robin-Everaars! Please fix the following issues with your PR:

- **Title**: Is too long (76 characters). The PR title must be strictly under 40 characters.

floatpanebot (dismissed):
Hi @Robin-Everaars! Please fix the following issues with your PR:

- **Title**: Is too long (42 characters). The PR title must be strictly under 40 characters.

Robin-Everaars:
I've seen the linting issues. I'll work on them later today (thought I could quickly open the PR in between other things).

andrinoff:
@Robin-Everaars this PR is way too big and will not backport to v1 automatically, requiring the team to remake this change on `release/v1` branch.

Please, remake this PR on `release/v1` branch

Robin-Everaars:
@andrinoff understood, I'll have a go at it sometime soon. Otherwise happy with the general direction?

Robin-Everaars:
And I'll close this PR to create several smaller, separate new PRs later

PR #1690, opened 2026-07-20
fix(imap): STARTTLS for non-993 ports

### LINKED ISSUE #1688: IMAP forces implicit TLS on any port except 1143/143 — breaks Proton Mail Bridge on non-default ports
matcha decides the IMAP TLS mode by hardcoded port number, not by config or server capability. It only uses STARTTLS when the port is 1143 or 143; on any other port it forces implicit TLS, which Proton Mail Bridge cannot serve → "folder fetch failed". Root cause: fetcher/fetcher.go connectWithOptions calls imapclient.DialTLS (unconditional implicit TLS) in the else branch for any port ≠ 1143/143. The SMTP path (sender/sender.go) correctly does STARTTLS on any non-465 port, so sending works but fetching fails. Full report: see attached file matcha_tls_bug_report.md

[matcha_tls_bug_report.md](https://github.com/user-attachments/files/30101781/matcha_tls_bug_report.md)

kawacukennedy:
## What?

Use STARTTLS for all non-993 IMAP ports, matching the SMTP path logic.

## Why?

Fixes #1688

Matcha previously only used STARTTLS on ports 1143 and 143, falling back to implicit TLS on every other port. This broke local IMAP bridges (e.g. Proton Mail Bridge) that serve IMAP on non-standard ports with STARTTLS. Port 993 uses implicit TLS; all other ports now correctly use STARTTLS.

### DIFF
--- a/fetcher/fetcher.go
+++ b/fetcher/fetcher.go
@@ -416,17 +416,24 @@ func connectWithOptions(account *config.Account, extraOpts *imapclient.Options) 
 	var c *imapclient.Client
 	var err error
 
-	// If using standard non-implicit ports (1143 or 143), use DialStartTLS
-	if imapPort == 1143 || imapPort == 143 {
-		c, err = imapclient.DialStartTLS(addr, options)
+	// Port 993 always uses implicit TLS. For all other ports, try STARTTLS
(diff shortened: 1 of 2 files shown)


kawacukennedy:

PR #1692, opened 2026-07-20
fix(sender): add SMTP dial timeouts

kawacukennedy:
## What?

Add 30-second connection timeouts to both SMTP dial paths (implicit TLS on port 465 and plain TCP for other ports).

## Why?

Both `tls.Dial` and `smtp.Dial` blocked indefinitely if the remote server was slow to respond, a firewall silently dropped packets, or DNS resolution stalled. This hung the entire application with no way to cancel.

Also fixed IPv6 address formatting (`fmt.Sprintf` → `net.JoinHostPort`).


floatpanebot (dismissed):
Hi @kawacukennedy! Please fix the following issues with your PR:

- **Title**: Is too long (49 characters). The PR title must be strictly under 40 characters.

kawacukennedy:

PR #1694, opened 2026-07-20
fix(config): timeout OAuth subprocesses

kawacukennedy:
## What?

Add context timeouts to both OAuth subprocess calls (30s for token retrieval, 5 min for browser flow). Capture stderr into a buffer instead of leaking it to `os.Stderr`.

## Why?

Both `exec.Command` calls suppressed the `noctx` linter and could hang indefinitely if the Python helper script stalled. Stderr was piped directly to `os.Stderr`, which could leak sensitive information (tokens, error traces) to logs.

### DIFF
--- a/config/oauth.go
+++ b/config/oauth.go
@@ -1,12 +1,15 @@
 package config
 
 import (
+	"bytes"
+	"context"
 	_ "embed"
 	"fmt"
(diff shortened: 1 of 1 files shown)


kawacukennedy:

PR #1701, opened 2026-07-24
fix(imap): oauth2 url fix for outlook

1unarzDev:
## What?

- Added try-catch block to make it easier to debug OAuth2 errors encountered while authenticating IMAP. 
- Revised Outlook URL generation code for updated Entra ID redirect

## Why?

I wasn't able to authenticate my Outlook emails previously. I'm not fully sure if the documentation clearly states that you need your Entra ID redirect to be configured as "Web" rather than "Mobile & Desktop App," but that is also a needed fix.


floatpanebot (dismissed):
Hi @1unarzDev! Please fix the following issues with your PR:

- **Title**: Is too long (67 characters). The PR title must be strictly under 40 characters.

andrinoff (approved):
lgtm

PR #1706, opened 2026-08-11
fix: skip update check on local builds

alejodelosrios:
## What?

`checkForUpdatesCmd` and `checkForV1RCCmd` now return early on a **local** development build — an unstamped binary running outside any package sandbox. Snap and Flatpak builds are explicitly left untouched.

## Why?

`version` defaults to `"dev"` (main.go:67), and `make build` / `go run .` build without ldflags (Makefile:25-26), so a source build is unstamped. `detectInstalledVersion()` then reports something that is never the running binary:

- With a packaged Matcha present, it falls back to the system package manager and returns *that* package's version — so the TUI advertises an update for a binary the package manager does not own.
- With no packaged Matcha at all, it returns `"dev"`, which never equals the latest release tag.

Either way the notice fires on every source build, e.g. `Update available: 0.44.0 (installed: dev)`.

The docs already frame source builds as contributor territory: "Building from source is not recommended for most users… Only proceed if you… want to contribute to the project or need a custom build" (docs/docs/installation.md).

## Why the guard needs two conditions, not one

Testing `version == "dev"` alone would regress released builds. `snapcraft.yaml` and `com.floatpane.matcha.yaml` both build without ldflags, so snap and flatpak binaries are *also* unstamped — and those **are** managed by the update path (`trySnapRefresh`, `tryFlatpakUpdate`). Their runtimes always inject `SNAP` and `FLATPAK_ID`, so the guard uses that to distinguish a packaged build from a local one.

| Build | stamped | sandbox | update check |
|---|---|---|---|
| `make build` / `go run .` | no | no | **skipped** (this change) |
| goreleaser / nix / `make install` | yes | no | unchanged |
| snap | no | yes | unchanged |
| flatpak | no | yes | unchanged |

This mirrors an existing pattern: `view/html.go:76-78` already uses `os.Getenv(...) != ""` in production to detect the runtime environment.

## Tests

Table-driven coverage of all seven cases, including `unstamped snap build` and `unstamped flatpak build`. Removing the sandbox condition fails exactly those two. `make lint`, `make test` and `go test -race` are clean.

## Known limitation

A developer running `go run .` from the terminal of a *packaged* editor (VS Code snap, or VS Code / GNOME Builder on Flatpak) inherits the editor's `SNAP` / `FLATPAK_ID` and will still see the notice. This fails conservative — showing the notice, not hiding it.

## Not included

`snapcraft.yaml` and the Flatpak manifest not stamping `main.version` is a real bug on its own (a released snap reports `dev`), but it is a separate logical change. Happy to open it as a follow-up.

### DIFF
--- a/main.go
+++ b/main.go
@@ -3717,6 +3717,26 @@ func downloadAttachmentCmd(account *config.Account, uid uint32, folderName strin
 	}
 }
 
+// isLocalDevBuild reports whether this binary is an unstamped build running
+// outside of any package sandbox, e.g. `make build` or `go run .` from a
+// clone. It matters for the update checks: detectInstalledVersion falls back
+// to asking the system package manager, so a contributor running their own
+// build is told the version of the *packaged* Matcha they happen to have
+// installed, which almost never matches the latest release tag.
+//
+// Version alone isn't a reliable signal: snap and flatpak builds are also
+// unstamped (snapcraft.yaml and the flatpak manifest build without ldflags),
+// but they *are* managed by the update path (trySnapRefresh, tryFlatpakUpdate).
+// SNAP and FLATPAK_ID are always injected by their respective runtimes
+// (snapenv.basicEnv; flatpak-run.c's "FLATPAK_ID is always set"), so their
+// presence rules out a local dev build even when the version is unstamped.
+func isLocalDevBuild() bool {
+	v := strings.TrimSpace(version)
+	unstamped := v == "" || v == "dev"
+	sandboxed := os.Getenv("SNAP") != "" || os.Getenv("FLATPAK_ID") != ""
+	return unstamped && !sandboxed
+}
+
 /*
 detectInstalledVersion returns a best-effort installed version string.
 Priority:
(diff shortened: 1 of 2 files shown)


floatpanebot (dismissed):
Hi @alejodelosrios! Please fix the following issues with your PR:

alejodelosrios:
Verified the sandbox case manually, since it's the part worth being sure about.

Same binary, same network, seconds apart — the only difference is one environment
variable:

```
./bin/matcha                          # unstamped, no sandbox → no notice
SNAP=/snap/matcha/x1 ./bin/matcha     # unstamped, sandboxed  → notice is back
```

The screenshot below is the second command (my locale is `es`, so the string is
translated): `Actualización disponible: 0.44.0 (instalada: dev)` — i.e. the update
check runs normally as soon as the binary looks packaged. A snap or flatpak user
keeps getting release notifications exactly as before; only a local source build
goes quiet.

The two table cases `unstamped snap build` / `unstamped flatpak build` cover this in
CI, and dropping the sandbox condition from the guard fails exactly those two and
nothing else.

[
Screenshot 2026-08-11 at 1 17 39 p m
](url)

PR #1716, opened 2026-09-11
fix(pgp): decrypt bare armored messages

### LINKED ISSUE #1715: PGP decryption always fails on server-side encrypted mail (raw armor passed to pgpmail.Read)
## Summary

PGP decryption of incoming mail fails with
`failed to decrypt PGP message: message: malformed MIME header line: -----BEGIN PGP MESSAGE-`
whenever the encryption was applied server-side (e.g. mailbox.org's "Encrypted Inbox" Sieve rule). Decryption of matcha-sent PGP mail works; server-encrypted mail never works. The root cause is in `decryptPGPMessage` in `fetcher/fetcher.go`.

## Environment

- matcha 0.44.0 (also reproduced on 0.39.0; still present in v1.0.0-rc2)
- macOS (Apple Silicon), Homebrew cask
- Provider: mailbox.org with "PGP encryption for incoming mails" (Sieve auto-encrypts inbound mail to the account's public key, RFC 3156 PGP/MIME)

## Reproduction

1. Enable mailbox.org Encrypted Inbox (Sieve encrypts inbound mail to your public key).
2. Receive any email. On the server it is stored as `multipart/encrypted` with an `application/octet-stream` part containing the ASCII-armored `-----BEGIN PGP MESSAGE-----` block.
3. Open the message in matcha. It always shows:

   ```
   PGP Decryption Failed: failed to decrypt PGP message: message: malformed MIME
   header line: -----BEGIN PGP MESSAGE-
   ```

4. The same message decrypts cleanly with `gpg --decrypt` using the same private key, so the key, passphrase, and ciphertext are fine.

## Root cause

`decryptPGPMessage` (fetcher/fetcher.go, ~L2091-2134) is called with the *bare armored payload* extracted from the `application/octet-stream` part (see the part-walk in `checkPart`, ~L1042-1102). It then passes those raw bytes straight to:

```go
mr, err := pgpmail.Read(bytes.NewReader(encryptedData), openpgp.EntityList{entityList[0]}, nil, nil)
```

`pgpmail.Read` (emersion/go-pgpmail v0.2.2) expects a **complete RFC 822 MIME entity** — it runs `textproto.ReadHeader` on the input first. The armor's first line `-----BEGIN PGP MESSAGE-----` contains no colon, so go-message returns `malformed MIME header line: -----BEGIN PGP MESSAGE-----` and decryption aborts before the private key is ever used.

## Suggested fix (either approach)

1. Armor path: if the payload starts with `-----BEGIN PGP MESSAGE-----`, decode with `openpgp.ReadArmoredMessage` and decrypt with `openpgp.Decrypt` directly, then feed the plaintext to `mail.CreateReader` (the caller already does this parse); or
2. MIME path: synthesize a minimal RFC 3156 wrapper around the extracted part (`multipart/encrypted` + `application/pgp-encrypted` version part + the octet-stream part) and pass that to `pgpmail.Read`.

## Why this matters

Server-side encryption (mailbox.org Sieve) is currently the *only* practical way to get encrypted mail at rest over open standards, since it requires no sender-side PGP. Any matcha user with mailbox.org's Encrypted Inbox hits this on every single message.

---

*Diagnosis and root-cause analysis assisted by Crush (CLI AI agent, model GLM by Z.ai). Reproduction environment and verification (gpg round-trip test on the same key pair) performed on the reporter's machine.*

chiliec:
## What?

`decryptPGPMessage` now handles a bare ASCII-armored payload: when the data starts with `-----BEGIN PGP MESSAGE-----` it decodes the armor and decrypts with `openpgp.ReadMessage` directly. Complete PGP/MIME entities still go through `pgpmail.Read` as before.

Adds `fetcher/pgp_test.go` covering both paths (bare armor + `multipart/encrypted` entity) using a throwaway key generated in the test.

## Why?

Fixes #1715. `checkPart` extracts the `application/octet-stream` part of a `multipart/encrypted` message and passes just that armored block to `decryptPGPMessage`, which handed it to `pgpmail.Read`. `pgpmail.Read` expects a full MIME entity and runs `textproto.ReadHeader` first, so the armor header line has no colon and decryption aborted with `malformed MIME header line: -----BEGIN PGP MESSAGE-----` before the private key was ever used. Server-side encrypted mail (mailbox.org Encrypted Inbox) hits this on every message.

Verification (Go 1.26.5, `GOTOOLCHAIN=local`):

- `go test ./fetcher/ -run TestDecryptPGP -v` — both tests pass.
- RED→GREEN: with the fix stashed, `TestDecryptPGPMessageBareArmor` fails with the exact reported error (`failed to decrypt PGP message: message: malformed MIME header line: -----BEGIN PGP MESSAGE-----`); `TestDecryptPGPMessageMIMEEntity` passes before and after, so the existing PGP/MIME path is unchanged.
- `go test ./fetcher/` — full package green; `go vet ./fetcher/`, `gofmt -l`, `golangci-lint run ./fetcher/` (v2.13.2, repo config) — 0 issues.

Not run locally: `./sender` tests (need `pkg-config`/pcsclite for the YubiKey dep on this box) — untouched by this change.


chiliec:
Note on the red `govulncheck` check: it reports stdlib advisories in go1.26.5 (GO-2026-6218 net/url, GO-2026-6090 crypto/tls, GO-2026-5972 encoding/asn1, fixed in 1.26.6) via call paths in `spellcheck`, `sender`, `main.go` and the existing IMAP dial code — none in this diff. The same job has been failing on `master` for the last three scheduled runs (Sep 7, Aug 31, Aug 24), so it's a toolchain bump, not this PR. Everything else (build/test on all 3 OSes, lint, gosec, e2e, fuzz) is green.

